PR #1, opened 2022-05-31
ci: clean ubuntu and python versions 

paulfouquet:
Remove multiple ubuntu and python versions to maintain


MDavidson17:
Reopening this PR as I accidentally deleted this branch, sorry.

pyproject.toml
```diff
@@ -25,22 +25,32 @@ profile = "black"
 [tool.mypy]
 show_error_codes = true
 strict = true
+disable_error_code = [
```
paulfouquet, line 28: What was the reason of disabling the errors related to `import`?

MDavidson17: No reason, I copied across some code from topo-procesor including this and the mypy overrides below.
I can remove it, and we can add it back as we go if its needed?

blacha (approved):

pyproject.toml
paulfouquet: Let's push it and do the cleaning in another PR

amfage (approved):

PR #11, opened 2022-06-13
docs: update readme(s)

MDavidson17:
- Overall repo readme.md updated to include information about the topo-imagery package. 
- readme.md in container folder deleted to reduce the risk of forgetting to update one or the other.

**Reasoning:**
https://docs.github.com/en/packages/learn-github-packages/publishing-a-package#about-published-packages

> "A repository can be connected to more than one package. To prevent confusion, make sure the README and description clearly provide information about each package."


amfage (approved):

README.md
```diff
@@ -14,3 +11,32 @@
 [![Imports: isort](https://img.shields.io/badge/%20imports-isort-%231674b1?style=flat&labelColor=2e3a44)](https://pycqa.github.io/isort/)
 [![Checked with mypy](http://www.mypy-lang.org/static/mypy_badge.svg)](http://mypy-lang.org/)
 [![Code Style: prettier](https://img.shields.io/badge/code_style-prettier-ff69b4.svg)](https://github.com/prettier/prettier)
+
+
+_A collection of packages used to process imagery_
+
+# Packages
```
paulfouquet, line 18: As we have only one package for now (and probably want only one for this repository for simplicity), we won't have a list of different containers.
What do you think of removing the `s` of `Packages` and not having a list?

README.md
```diff
@@ -14,3 +11,32 @@
...
+## [1. topo-imagery](https://github.com/linz/topo-imagery/tree/master/containers/gdal)
+
+### GDAL Container
```
paulfouquet, line 21: Same here, should we rename this title as `Container description` or something like that?

paulfouquet (changes requested):
As per our discussion earlier during standup, we might only have one container in this repo. I've got a couple of suggestions around that.

paulfouquet (changes requested):
I would suggest to use the type `docs` instead of `fix` for this PR
> docs: Documentation only changes

[*source](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)

MDavidson17:
> I would suggest to use the type `docs` instead of `fix` for this PR
> 
> > docs: Documentation only changes
> 
> [*source](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary)

good point, very true :+1:

paulfouquet (approved):

amfage (approved):

PR #12, opened 2022-06-19
ci: add versioning

MDavidson17:
- version bump script added
- docker publishing script updated
 
**note: I have been unable to test this code in full so all comments and theories are appreciated**


paulfouquet (dismissed):

version.bump.sh
```diff
@@ -0,0 +1,29 @@
...
+echo v${CURRENT_VERSION} | tee VERSION
+
+# Commit the changed files
+git commit -a --amend --no-edit
+
+# Checkout a new release branch
+git checkout -b release/v${CURRENT_VERSION}
+
+# This tag will be created once the pull request is merged
+git tag -d v${CURRENT_VERSION}
```
blacha, line 29: might just be a option to push the version tag directly, is there a reason we want a pull request based versioning scheme here?

PR #23, opened 2022-06-21
feat: add standardising script (TDE-303)

paulfouquet:
## Description
Add the `standardising` script

## Changes

- `standardising.py` has been added in the `script` directory
- Unit tests have been initialized
- some changes in `aws_helper` related to assuming a role

### DIFF
--- a/scripts/aws_helper.py
+++ b/scripts/aws_helper.py
@@ -1,5 +1,8 @@
 import json
+import os
 from os import environ
+from typing import Tuple
+from urllib.parse import urlparse
 
 import boto3
 from linz_logger import get_log
@@ -34,7 +37,7 @@ def init_roles():
     for cfg in json_content["buckets"]:
         bucket_roles[cfg["bucket"]] = cfg
 
-def get_credentials(bucket_name: str):
+def get_credentials(bucket_name: str, set_credentials: bool = False):
     get_log().debug("get_credentials_bucket_name", bucket_name=bucket_name)
     if not bucket_roles:
         init_roles()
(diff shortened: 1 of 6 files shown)


scripts/aws_helper.py
```diff
@@ -49,6 +52,10 @@ def get_credentials(bucket_name: str):
                 assumed_role_object["Credentials"]["SecretAccessKey"],
                 assumed_role_object["Credentials"]["SessionToken"],
             )
+        if set_credentials:
```
blacha, line 55: We should never overwrite the system environment variables.

You should take these credentials and add them into the environment of the command

scripts/standardising.py
```diff
@@ -0,0 +1,45 @@
...
+src_bucket_name, src_file_path = aws_helper.parse_path(source)
+dst_bucket_name, dst_path = aws_helper.parse_path(destination)
+get_log().debug("source", bucket=src_bucket_name, file_path=src_file_path)
+get_log().debug("destination", bucket=dst_bucket_name, file_path=dst_path)
+dst_bucket = aws_helper.get_bucket(dst_bucket_name)
+
+with tempfile.TemporaryDirectory() as tmp_dir:
+    standardized_file_name = f"standardized_{os.path.basename(src_file_path)}"
+    tmp_file_path = os.path.join(tmp_dir, standardized_file_name)
+    src_gdal_path = f"/vsis3/{source.replace('s3://', '')}"
```
blacha, line 28: replce the `s3://` with `/vsis3` this will let it run on files that are not in s3.

paulfouquet: My bad, stupid copy/paste from `Topo-Processor`...

scripts/aws_helper.py
paulfouquet: Yes, I was concerned by removing the default credentials by doing this.

amfage (approved):

PR #25, opened 2022-06-22
ci: formatting (TDE-400)

MDavidson17:
runs: black, isort, pylint, gitlint and mypy (using .pre-commit-config.yaml)
on: push and pull request


scripts/create_polygons.py
```diff
@@ -4,20 +4,21 @@
 from collections import Counter
 from urllib.parse import urlparse
 
-import aws_helper as aws_helper
+import aws_helper
 from linz_logger import get_log
-from osgeo import gdal
+from osgeo import gdal  # pylint: disable=import-error
```
paulfouquet, line 9: Maybe we should add a comment here saying that this library is not imported with `Poetry` because of import issues, but included in the Docker image we use.

amfage (dismissed):

scripts/aws_helper.py
```diff
@@ -1,11 +1,12 @@
 import json
 from os import environ
+from typing import Any
 
 import boto3
 from linz_logger import get_log
 
 
-class Credentials:
+class Credentials:  # pylint: disable=too-few-public-methods
```
paulfouquet, line 9: Not necessary for this PR (and I've created this class...), but it would be better to use a `namedtuple` instead of a `class`. Feel free to do it or not.

paulfouquet (changes requested):
A couple of suggestions

paulfouquet (dismissed):

paulfouquet (approved):

amfage (approved):

PR #36, opened 2022-07-01
feat: enable list source TDE-411

MDavidson17:
`nargs="+"` added to --source to enable multiple input files.

ngars puts all inputs into a list:
One input: `--source image.tiff` -> `[image.tiff]`
Multiple Inputs: `--source image_one.tiff image_two.tiff` -> `[image_one.tiff, image_two.tiff]`
Using basemaps cli: `--source "[\"image_one.tiff\", \"image_two.tiff\"]"` -> `["[\"image_one.tiff\", \"image_two.tiff\"]"]`

Only the basemaps cli needs to be formatted
`["[\"image_one.tiff\", \"image_two.tiff\"]"]` ->  `[image_one.tiff, image_two.tiff]`


scripts/format_source.py
```diff
@@ -0,0 +1,11 @@
+import json
+from typing import List, cast
+
+
+def format_source(source: List[str]) -> List[str]:
+    """If source from basemaps cli it has a string which contains a list that needs to be split
```
blacha, line 6: this is actually a Argo constraint that baseamaps, Argo only passes JSON objects into sub commands 😭

blacha (dismissed):

amfage (approved):

MDavidson17:
Code succesfully runs in argo
image

scripts/format_source.py
```diff
@@ -0,0 +1,11 @@
+import json
+from typing import List, cast
+
+
+def format_source(source: List[str]) -> List[str]:
+    """Due to Argo constraints if using the basemaps cli the source has a string that contains a list that needs to be split.
```
paulfouquet, line 6: Should we specify basemaps cli **list** command?

amfage (approved):

scripts/format_source.py
```diff
@@ -0,0 +1,11 @@
...
+    example: ["[\"s3://test/image_one.tiff, \"s3://test/image_two.tiff\"]"]
+    """
+    if len(source) == 1 and source[0].startswith("["):
+        return cast(List[str], json.loads(source[0]))
```
paulfouquet, line 10: I would `json.loads` anyways return the string if `loads` raises an exception. In your case, if `source[0]` is a `str` like "[my_string" it would fails. Or maybe we want it to fails, but it would be good to nicely catch the exception in that case.

scripts/tests/format_source_test.py
```diff
@@ -0,0 +1,32 @@
+from format_source import format_source
+
+
```
paulfouquet, line 3: I would suggest adding a test when the format is not `json` "compatible"

paulfouquet (changes requested):
a couple of comments

scripts/format_source.py
```diff
@@ -0,0 +1,17 @@
...
+
+def format_source(source: List[str]) -> List[str]:
+    """Due to Argo constraints if using the basemaps cli list command
+    the source has a string that contains a list that needs to be split.
+    example: ["[\"s3://test/image_one.tiff\", \"s3://test/image_two.tiff\"]"]
+    """
+    if len(source) == 1 and source[0].startswith("["):
+        try:
+            return cast(List[str], json.loads(source[0]))
+        except ValueError as e:
```
paulfouquet, line 15: I would specify the [JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError) so you're not bypassing another issue where `ValueError` could be more broad

scripts/format_source.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def format_source(source: List[str]) -> List[str]:
+    """Due to Argo constraints if using the basemaps cli list command
+    the source has a string that contains a list that needs to be split.
+    example: ["[\"s3://test/image_one.tiff\", \"s3://test/image_two.tiff\"]"]
+    """
+    if len(source) == 1 and source[0].startswith("["):
+        try:
+            return cast(List[str], json.loads(source[0]))
```
paulfouquet, line 14: Was `mypy` complaining so you had to add `cast`? which is actually not casting the Python object returned by `json.loads()` but used for `typing`. In our case the json is an array so `json.loads` returns a Python `list`

paulfouquet (approved):

amfage (approved):

PR #40, opened 2022-07-07
feat: Non Visual QA script (TDE-309)

paulfouquet:
## Description
Cloudify the [non visual QA script](https://github.com/linz/aerial-imagery-scripts/blob/master/non_visual_qa/non_visual_QA.py)

## Future improvement
- explore ways of providing/displaying the output - current ouput in logs:
```
[2022-07-07T04:58:27.074Z] INFO non_visual_qa_result result={ 0=noDataValue not set 1=not 3 bands, 4 bands found 2=different srs 3=unexpected color interpretation bands; band 4 Undefined 4=b'Warning 1: The definition of projected CRS EPSG:2193 got from GeoTIFF keys is not the same as the one from the EPSG registry, which may cause issues during reprojection operations. Set GTIFF_SRS_SOURCE configuration option to EPSG to use official parameters (overriding the ones from GeoTIFF keys), or to GEOKEYS to use custom values from GeoTIFF keys and drop the EPSG code.\n' }

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,3 +21,5 @@ COPY ./scripts/create_polygons.py /app/
 COPY ./scripts/standardising.py /app/
 COPY ./scripts/aws_helper.py /app/
 COPY ./scripts/gdal_helper.py /app/
+COPY ./scripts/file_helper.py /app/
+COPY ./scripts/non_visual_qa.py /app/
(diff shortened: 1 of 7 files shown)


MDavidson17 (approved):
:+1:

PR #53, opened 2022-07-14
fix: skip files which are not tiff

paulfouquet:


scripts/create_polygons.py
```diff
@@ -51,6 +52,9 @@ def main() -> None:  # pylint: disable=too-many-locals
     output_files = []
 
     for file in source:
+        if not is_tiff(file):
+            get_log().error("create_polygon_file_not_tiff_skipped", file=file)
```
MDavidson17, line 56: Should these be info/debug rather than an error. 
I'm not sure but I feel like we'll be expecting to skip a number of files :thinking:

paulfouquet: Right, even `trace` would be best

amfage (dismissed):

blacha (approved):

blacha (approved):

amfage (approved):

PR #55, opened 2022-07-15
fix: log Non Visual QA error report in a JSON format

paulfouquet:
## Description
In order to facilitate the loggin in Elastic Search, the Non Visual QA error report should be in a JSON format.

## Changes

- Create a `dictionnary`
```json
{"file": "path/to/the/file", "errors": [{"type": "error_type", "message": "error_message"}]}
```

## Example
```json
{"file": "s3://linz-imagery-staging/RGBi4/tasman_rural_2020_0.3m_a_RGBI/tif/2020_BQ24_5000_0601.tif", "errors": [{"type": "gdalinfo", "message": "error(s): b'Warning 1: The definition of projected CRS EPSG:2193 got from GeoTIFF keys is not the same as the one from the EPSG registry, which may cause issues during reprojection operations. Set GTIFF_SRS_SOURCE configuration option to EPSG to use official parameters (overriding the ones from GeoTIFF keys), or to GEOKEYS to use custom values from GeoTIFF keys and drop the EPSG code.\\n'"}, {"type": "nodata", "message": "noDataValue not set"}, {"type": "bands", "message": "bands count is not 3", "count": "4"}, {"type": "color", "message": "unexpected color interpretation bands", "missing": "band 4 Undefined"}, {"type": "srs", "message": "different srs"}]}
```

### DIFF
--- a/scripts/gdal_helper.py
+++ b/scripts/gdal_helper.py
@@ -6,6 +6,10 @@ from aws_helper import get_bucket_name_from_path, get_credentials, is_s3
 from linz_logger import get_log
 
 
+class GDALExecutionException(Exception):
+    pass
+
+
 def get_vfs_path(path: str) -> str:
     """Make the path as a GDAL Virtual File Systems path.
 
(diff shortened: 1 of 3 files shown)


scripts/gdal_helper.py
```diff
@@ -67,8 +71,8 @@ def run_gdal(command: List[str], input_file: str = "", output_file: str = "") ->
             capture_output=True,
         )
     except subprocess.CalledProcessError as cpe:
-        get_log().error("run_gdal_failed", command=command_to_string(command))
-        raise cpe
+        get_log().error("run_gdal_failed", command=command_to_string(command), error=str(cpe.stderr, "utf-8"))
+        raise GDALExecutionException(f"GDAL {str(cpe.stderr, 'utf-8')}") from cpe
     get_log().debug("run_gdal_translate_succeded", command=command_to_string(command))
```
blacha, line 76: should check stderr for any warnings/issues too.

also stdout might be interesting to dump as a log message?

paulfouquet: > should check stderr for any warnings/issues too.
Yes, it is actually checked in the script itself but it should be done here you're right.

> also stdout might be interesting to dump as a log message?
:+1:

amfage (approved):

PR #69, opened 2022-07-26
feat: File System Abstraction local/AWS TDE-407

paulfouquet:
## Description
This is a start to build a File System Abstraction in Topo-Imagery container. This change focus on having a `read` and `write` method which abstract the fact that the files are local or on AWS S3.

## Changes
- `fs` provides a `read` and a `write` methods which allows local or AWS S3 path.
- A bit of refactoring on the python project structure.


scripts/aws/aws_helper.py
```diff
@@ -79,6 +79,14 @@ def get_bucket_name_from_path(path: str) -> str:
 
 
 def parse_path(path: str) -> Tuple[str, str]:
```
blacha, line 81: A tuple doesnt say what is what, so using something that knows that the first part is `bucket` would be easier for people to understand

```python
ret = parse_path("s3://foo/bar")
ret.bucket # foo
ret.key # /bar
```

scripts/files/fs_local.py
```diff
@@ -0,0 +1,22 @@
+def write(destination: str, source: bytes) -> None:
+    """Write the source to the local destination file.
+
+    Args:
+        destination (str): The local path to the file to write.
+        source (bytes): The source file in bytes.
+    """
+    with open(destination, "wb") as buffer:
```
blacha, line 8: NIT: You are opening a `file` not a `buffer`.

scripts/files/fs_s3.py
```diff
@@ -0,0 +1,101 @@
...
+    Args:
+        destination (str): The AWS S3 path to the file to write.
+        source (bytes): The source file in bytes.
+        needs_credentials (bool, optional): Tells if credentials are needed. Defaults to False.
+    """
+    if source is None:
+        get_log().error("write_s3_source_none", path=destination, error="The 'source' is None.")
+        raise Exception("The 'source' is None.")
+    bucket_name, key = parse_path(destination)
+    key = key[1:]
```
blacha, line 46: this is a odd statement why is this needed?

scripts/files/fs_s3.py
```diff
@@ -0,0 +1,101 @@
...
+    s3 = boto3.resource("s3")
```
blacha, line 47: You could simplify this to always calling `get_auth_s3_session(bucket_name, "write")` which then only gives you a authed s3 object if you need it.

scripts/files/fs_s3.py
```diff
@@ -0,0 +1,101 @@
...
+def get_auth_s3_session(bucket_name: str) -> S3ServiceResource:
+    """Return a boto3 S3 Resource with the AWS credentials for a bucket.
+
+    Args:
+        bucket_name (str): The name of the bucket.
+
+    Returns:
+        S3ServiceResource: The boto3 S3 Resource.
+    """
+    credentials = get_credentials(bucket_name)
```
blacha, line 24: you could try reading/writing the location before creating credentials as the default creds may have enough power to read or write.

scripts/files/fs_s3.py
```diff
@@ -0,0 +1,101 @@
...
+    session = boto3.Session(
```
blacha, line 25: I think it would be better to use a specific AssumeRoleCredential helper similar to how it is done https://github.com/linz/lds-cache-client-py/blob/master/lds_cache/lds_cache_source.py#L35

scripts/files/fs_s3.py
paulfouquet: That's what it's trying to do in the `read` or `write` methods actually

scripts/files/fs_s3.py
paulfouquet: To remove the leading `/`. This will be put in the `parse_path` method instead.

scripts/files/fs_s3.py
paulfouquet: I leave the refreshable credential for another ticket (TDE-235)

scripts/files/fs_s3.py
paulfouquet: I leave that for when/if `write` roles will be exposed.

scripts/aws/aws_helper.py
```diff
@@ -13,6 +13,7 @@
     Bucket = object
 
 Credentials = namedtuple("Credentials", ["access_key", "secret_key", "token"])
+S3_path = namedtuple("S3_path", ["bucket", "key"])
```
blacha, line 16: Why named tuples? do you have reasoning for using them vs other container classes?

Also mixing Snake and TitleCase is a little odd,  what is the prefered method for naming Objects Title?

scripts/converters/format_source.py
```diff
@@ -0,0 +1,32 @@
...
+    """Due to Argo constraints if using the basemaps cli list command
+    the source has a string that contains a list that needs to be split.
+    example: ["[\"s3://test/image_one.tiff\", \"s3://test/image_two.tiff\"]"]
+    """
+    if len(source) == 1 and source[0].startswith("["):
+        try:
+            source_json: List[str] = json.loads(source[0])
+            return source_json
+        except json.JSONDecodeError as e:
+            get_log().debug("Decoding Json Failed", source=source, msg=e)
```
blacha, line 18: should this really swallow the error? If a invalid input is detected should it not throw?

scripts/create_polygons.py
```diff
@@ -46,43 +47,37 @@ def main() -> None:
...
+                    get_log().debug(f"- data_px_count was zero in create_mask function for the tif {mask_file}")
+                else:
+                    destination_file_name = os.path.splitext(source_file_name)[0] + ".geojson"
+                    temp_file_path = os.path.join(tmp_dir, destination_file_name)
+                    polygonize_command = f'gdal_polygonize.py -q "{mask_file}" "{temp_file_path}" -f GeoJSON'
+                    os.system(polygonize_command)
+
+                output_files.append(temp_file_path)
+        except ClientError as ce:
+            get_log().error("create_polygon_skip_file", error=str(ce))
```
blacha, line 79: we should fail the command if something fails?

scripts/files/fs_s3.py
```diff
@@ -0,0 +1,100 @@
...
+        if needs_credentials:
+            s3 = _get_s3_resource(s3_path.bucket)
+
+        s3_object = s3.Object(s3_path.bucket, key)
+        file: bytes = s3_object.get()["Body"].read()
+    except botocore.exceptions.ClientError as ce:
+        if not needs_credentials and ce.response["Error"]["Code"] == "AccessDenied":
+            get_log().debug("read_s3_needs_credentials", path=path)
+            return read(path, True)
+        if ce.response["Error"]["Code"] == "NoSuchBucket":
```
blacha, line 92: you should use the `except boto...exceptions....NoSuchBucket` style for this

scripts/aws/aws_helper.py
paulfouquet: > Why named tuples? do you have reasoning for using them vs other container classes?

1. I was using a `tuple` before and did not want to create a `class` when the purpose was only to name those properties.
2. I found this interesting topic, https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases

> Also mixing Snake and TitleCase is a little odd, what is the prefered method for naming Objects Title?

This is purely a mistake, this should be `s3_path`, good spotting.

scripts/converters/format_source.py
paulfouquet: The way this change appears is really confusing. I haven't write this code. I just moved the file.

scripts/create_polygons.py
paulfouquet: This one is interesting. Let's say it fails on reading the file from its source which is a s3 bucket. If for some reasons, some other files are on another bucket that we have access too, should we stop them to be processed?

scripts/aws/aws_helper.py
blacha: Just looking at the articile you may want `NamedTuple` then

```python
from collections import namedtuple
from typing import NamedTuple

Foo = namedtuple("Foo", ["a", "b"])
Foo("a","b").a # Any
Foo("a", 1).b # Any

Bar = NamedTuple("Bar", [("a", str), ("b", int)])
Bar("a", 1).a # str
Bar("a", 1).b # int
```

scripts/files/fs_s3.py
paulfouquet: Ha, I removed my comment somehow. When using boto3 `Resources`, exceptions are a bit different than for `Client`. 
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#catching-exceptions-when-using-a-resource-client
I am not able to find the `AccessDenied` exception. Should I keep the existing behavior for `AccessDenied` and catch the exception using `except` on each specific exceptions for the other errors?

scripts/aws/aws_helper.py
paulfouquet: I should specify the type and use TitleCase then, if it is the convention

scripts/create_polygons.py
blacha: I would process them all and just fail the command at the end

paulfouquet: Meaning `continue` to the next file when one fails, keep the information that it failed and at the end fail the command if at least one failed?

scripts/converters/format_source.py
paulfouquet: I'll fix that in an other PR

scripts/create_polygons.py
paulfouquet: This should not be part of this PR

blacha: this should be part of the PR, this PR changes the logic from exiting properly with a error code if something goes wrong to swallowing the errors.

MDavidson17 (approved):

amfage (approved):

PR #77, opened 2022-08-03
feat: logs hostname and execution time TDE-431

paulfouquet:
## Description
To facilitate logs investiation, having a unique ID as `hostname` and a `duration` for the scripts execution will help.

## Changes
- Upgrade [`python_linz_logger`](https://github.com/linz/python-linz-logger) to `v0.9.0` to get `hostname` as a default key.
- Capture the execution time (`duration`) and log it.

## Tests
The execution time has been tested locally. Getting the Argo node ID has been tested using the `test.yaml` Workflow.


scripts/create_polygons.py
```diff
@@ -41,14 +42,17 @@ def get_pixel_count(file_path: str) -> int:
...
+    start_time = time_in_ms()
+    node_id = os.getenv("ARGO_NODE_ID")
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--source", dest="source", nargs="+", required=True)
     arguments = parser.parse_args()
     source = arguments.source
 
     source = format_source(source)
+    get_log().info("create_polygons_start", nodeId=node_id, source=source)
```
blacha, line 54: `nodeId` needs to be added to *every* log message not just the first one. I think there is something like `get_log().add_keys(?)`

I wonder if it is also just the `hostname` key that is common to most of our other logs?

paulfouquet: Yes my bad. `get_log().add_keys()` allows you to permanently add a key to your logger instance?

blacha (approved):

PR #92, opened 2022-08-15
fix: TDE-454 remove `standardise` from file name

MDavidson17:
So that we can get the stac item id from the filename

### DIFF
--- a/scripts/standardising.py
+++ b/scripts/standardising.py
@@ -40,7 +40,7 @@ def standardising(file: str) -> str:
     get_log().info("standardising_start", source=file)
 
     _, src_file_path = parse_path(file)
-    standardized_file_name = f"{get_file_name_from_path(s3_path.key)}.tiff"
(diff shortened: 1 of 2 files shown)


scripts/standardising.py
```diff
@@ -23,7 +23,7 @@
         continue
 
     s3_path = parse_path(file)
-    standardized_file_name = f"standardized_{get_file_name_from_path(s3_path.key)}"
+    standardized_file_name = f"{get_file_name_from_path(s3_path.key)}"
```
paulfouquet, line 26: You don't need to format the string (`f" "`) as `get_file_name_from_path()` returns a string

paulfouquet (changes requested):

blacha (approved):

paulfouquet (approved):

amfage (approved):

PR #96, opened 2022-08-19
feat: gdal concurrency TDE-457

amfage:


scripts/standardising.py
```diff
@@ -1,80 +1,100 @@
...
 from scripts.aws.aws_helper import parse_path
-from scripts.cli.cli_helper import parse_source
+from scripts.cli.cli_helper import is_argo, parse_source
 from scripts.files.files_helper import get_file_name_from_path, is_tiff
 from scripts.gdal.gdal_helper import run_gdal
 from scripts.logging.time_helper import time_in_ms
 
 
-def standardising(files: List[str]) -> List[str]:
+def start_standardising(files: List[str], argo_env: bool) -> List[str]:
```
blacha, line 14: would be good to do something like `cryptocurrency: int` and specify the number of workers to use, as `argo_env` doesn't really make too much sense to the caller.

amfage: Thanks, could you elaborate further on `cryptocurrency` please?

blacha: LOL darn autocorrect! `concurrency` 

so then this just takes the number of threads rather than a opaque `is_in_argo`

scripts/aws/aws_helper.py
```diff
@@ -85,7 +85,7 @@ def parse_path(path: str) -> S3Path:
         path (str): A S3 path.
 
     Returns:
-        S3Path (NamedTupe): s3_path.bucket (str), s3_path.key (str)
```
blacha, line 88: not 100% related to this, but does vscode not detect automatically the types of all of these, we shouldnt be doubling up the typing in both the function definition and the docs as it often leads to them being out of sync.

scripts/standardising.py
```diff
@@ -1,80 +1,102 @@
...
-            "-co",
-            "overview_quality=90",
-            "-co",
-            "sparse_ok=true",
-        ]
-        run_gdal(command, input_file=file, output_file=tmp_file_path)
-        output_files.append(tmp_file_path)
+
+    if concurrency:
+        with Pool(concurrency) as p:
```
blacha, line 28: is there any down sides to doing `with Pool(1) as p` instead of doing two different code paths for running the script?

amfage: I didn't want to add any overhead for local file processing, but I could test to see if it is an issue.

amfage: It does not seem to be an issue.

blacha (approved):

PR #100, opened 2022-08-22
feat: TDE-452 create imagery stac item

MDavidson17:
- stac item is created based on a [similar implementation in basemaps](https://github.com/linz/lds-cache/blob/master/src/stac.ts#L10)
- ~~pystac was only used to get the stac version, this can be hardcoded and pystac removed entirely, or the code can be changed to return a pystac item object.~~ -pystac removed

nb: The code has been implemented with thought of how the `create_stac_items.py` script might run for different datasets.
How it is currently written it could be possible to add a datatype dropdown input in the argo workflow and use this datatype to determine what stac to create. 
In topo-processor this was premtively scripted, however on reflection the developers mentioned they wouldn't have implemented it until there were additional datasets, therefore it has not been implemented yet in topo-imagery.


scripts/stac/imagery_stac.py
```diff
@@ -0,0 +1,68 @@
+from typing import Any, Dict, List
+
+from pystac import get_stac_version
+
+from scripts.files.files_helper import get_file_name_from_path, strip_extension
+from scripts.stac.util import checksum  # dont touch this
```
paulfouquet, line 6: Could you add an explaination on why not touching that? and maybe some additional research should be done to clearly understand the issue

scripts/stac/imagery_stac.py
```diff
@@ -0,0 +1,68 @@
+from typing import Any, Dict, List
+
+from pystac import get_stac_version
```
paulfouquet, line 3: Importing a library to get the stac version might be a bit too extrem

scripts/stac/imagery_stac.py
MDavidson17: ha, oops I forgot about this comment, I didn't mean for it to come through, but yes I will update it.

scripts/create_stac_items.py
```diff
@@ -0,0 +1,52 @@
...
+    arguments = parser.parse_args()
+
+    files = format_source(arguments.source)
+    date = format_date(arguments.date)
+
+    get_log().info("create_stac_items_start", source=files)
+
+    create_imagery_items(files, date)
+
+    get_log().info("create_stac_items_complete", source=files, duration=time_in_ms() - start_time)
```
paulfouquet, line 48: I would put the logs in `create_imagery_items` method

scripts/files/files_helper.py
```diff
@@ -5,5 +5,11 @@ def get_file_name_from_path(path: str) -> str:
     return os.path.basename(path)
 
 
+def strip_extension(file: str) -> str:
```
paulfouquet, line 8: You should also take account of `.TIF` and `.TIFF`
To avoid a `if` I would suggest doing something like:
```python
>>> extension = ".tif"
>>> stripped = "my_file.TIFF".lower().split(extension, 1)[0]
>>> print(stripped)
my_file
```

scripts/gdal/gdalinfo.py
```diff
@@ -0,0 +1,26 @@
...
+from scripts.gdal.gdal_helper import GDALExecutionException, run_gdal
+
+
+def gdal_info(path: str) -> Dict[Any, Any]:
+    gdalinfo_command = ["gdalinfo", "-stats", "-json"]
+
+    try:
+        gdalinfo_process = run_gdal(gdalinfo_command, path)
+        gdalinfo_result = {}
+        try:
```
paulfouquet, line 15: What is the need of having a nested `try/except` here?

scripts/files/files_helper.py
MDavidson17: I believe that with the changes I've added to https://github.com/linz/topo-imagery/pull/92 this function can be deleted, I'll do that once that PR is merged and I can rebase

scripts/gdal/gdalinfo.py
MDavidson17: Uhh, I was trying to replicate the gdalinfo code in non_visual_qa.py
If its unnecessary I can remove it.

scripts/cli/cli_helper.py
```diff
@@ -26,12 +28,46 @@ def parse_source() -> List[str]:
...
+    arguments = parse.parse_args()
 
     return format_source(arguments.source)
 
 
 def is_argo() -> bool:
     return bool(environ.get("ARGO_TEMPLATE"))
+
+
+def format_date(date: datetime) -> str:
```
paulfouquet, line 42: It would be a good idea to have some unit tests for these formatters and converters

scripts/gdal/gdalinfo.py
paulfouquet: OK, my bad, I'll refactor that and in `non_visual_qa.py` in another way later. That's ok for now.

amfage (changes requested):

MDavidson17:
this work has been superceeded by:
https://github.com/linz/topo-imagery/pull/133
https://github.com/linz/topo-imagery/pull/132
https://github.com/linz/topo-imagery/pull/134
https://github.com/linz/topo-imagery/pull/135

PR #108, opened 2022-08-24
feat: switch between two gdal presets "lzw" or "webp"

blacha:
Both are lossless compression for the top level imagery.

Webp is best suited for when there is RGB or RGBA imagery, LZW is best suited for any TIFF that is not Webp.

### DIFF
--- a/scripts/gdal/gdal_preset.py
+++ b/scripts/gdal/gdal_preset.py
@@ -1,5 +1,7 @@
 from typing import List
 
+from linz_logger import get_log
+
 GDAL_PRESET_LZW = [
     "gdal_translate",
     "-q",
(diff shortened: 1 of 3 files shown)


scripts/gdal/gdal_preset.py
```diff
@@ -0,0 +1,76 @@
...
+    "blocksize=512",
+    "-co",
+    "overview_quality=90",
+    "-co",
+    "sparse_ok=true",
+]
+
+
+def get_gdal_command(preset: str) -> List[str]:
+    print(preset)
```
paulfouquet, line 71: is it a leftover from your tests? should we log instead?

blacha: oops, yeah should log it

paulfouquet (approved):

paulfouquet (approved):

PR #116, opened 2022-08-29
fix: not logging source as a list of string

paulfouquet:
## Description
We should not log the `source` as `list` of `str`: it can be very long, and also Elastic Search is not parsing it correctly.

## Change
Remove logging the full source and keeping only logging single file path when needed.

### DIFF
--- a/scripts/cli/cli_helper.py
+++ b/scripts/cli/cli_helper.py
@@ -5,6 +5,8 @@ from typing import List
 
 from linz_logger import get_log
 
+from scripts.logging.formatter import list_to_str
+
 
 def format_source(source: List[str]) -> List[str]:
     """Due to Argo constraints if using the basemaps cli list command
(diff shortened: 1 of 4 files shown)


scripts/cli/cli_helper.py
```diff
@@ -16,7 +18,7 @@ def format_source(source: List[str]) -> List[str]:
             source_json: List[str] = json.loads(source[0])
             return source_json
         except json.JSONDecodeError as e:
-            get_log().debug("Decoding Json Failed", source=source, msg=e)
+            get_log().debug("Decoding Json Failed", msg=e)
```
blacha, line 21: should this throw ?

paulfouquet: https://github.com/linz/topo-imagery/pull/117

blacha (approved):

PR #117, opened 2022-08-29
fix: raise an exception when parsing source encounters a JSONDecode TDE-448

paulfouquet:


scripts/standardising.py
```diff
@@ -59,7 +59,11 @@ def main() -> None:
     parser.add_argument("--source", dest="source", nargs="+", required=True)
     arguments = parser.parse_args()
 
-    source = format_source(arguments.source)
+    try:
+        source = format_source(arguments.source)
+    except Exception as e:
```
blacha, line 64: what happens if we dont catch it here? its already logged, so the CLI will just die right? what is the desired behaviour?

paulfouquet: Probably not throwing an exception to the user :)

scripts/non_visual_qa.py
```diff
@@ -152,7 +152,7 @@ def non_visual_qa(files: List[str]) -> None:
     get_log().info("non_visual_qa_end", source=files, duration=time_in_ms() - start_time)
 
 
-def main() -> None:
+def main() -> None:  # pylint: disable=duplicate-code
```
paulfouquet, line 155: That's annoying that pylint does not raise this warning locally via `pre-commit` script

scripts/standardising.py
blacha: I think a bunch of that main function could throw straight back to the client.

I would generally have something like

```python

try:
  main()
except Exception as e:
  log_exception(e)
  exit(1)
```

where your logging all exceptions, and then exiting, how does `argparse` suggest you do error handling in its CLIs?

paulfouquet: > how does argparse suggest you do error handling in its CLIs?

[argparse Invalid arguments](https://docs.python.org/3/library/argparse.html#invalid-arguments)
Looks like they do something similar?

amfage:
This is now covered in https://github.com/linz/topo-imagery/pull/582

PR #119, opened 2022-08-29
fix: accept preset arg standardise-validate

amfage:


scripts/standardising.py
```diff
@@ -53,18 +52,9 @@ def standardising(file: str, preset: str) -> str:
...
-    arguments = parser.parse_args()
 
-    source = format_source(arguments.source)
-
-    if is_argo():
-        concurrency = 4
-
-    start_standardising(source, arguments.preset, concurrency)
+    source = parse_source()
+    start_standardising(source, "lzw", 1)
```
blacha, line 57: shouldn't this also be able to run with webp and mutlithreaded?

amfage: I think we agreed to revisit that when there was a need to run standardise alone. Otherwise mypy complains about code duplication.

paulfouquet: mypy or pylint?

blacha: is it time to disable that lint rule? seems very onerous

paulfouquet: another approach would be to allow x lines of duplicated code

blacha: not allowing 4 lines of duplicated code seems like a overly strict rule?

amfage: I've reverted the `standardising` script to take arguments, `pylint` seems happy so :shrug:

blacha (approved):

PR #121, opened 2022-08-30
fix: suppress aux xml file creation

amfage:


scripts/non_visual_qa.py
```diff
@@ -94,7 +94,7 @@ def check_color_interpretation(self, gdalinfo: Dict[str, Any]) -> None:
 
     def run(self) -> None:
         gdalinfo_success = True
-        gdalinfo_command = ["gdalinfo", "-stats", "-json"]
+        gdalinfo_command = ["gdalinfo", "-stats", "-json", "--config", "GDAL_PAM_ENABLED", "NO"]
```
blacha, line 97: Could we add a comment as to what a PAM is?  and what this is trying to achieve? 

Does this also need to happen to the `gdal_translate` ?

amfage: I have added a comment. The auxiliary file is not created when `gdal_translate` is run.

MDavidson17 (approved):

PR #132, opened 2022-09-09
Feat: initialise collection stac TDE-453

MDavidson17:
1

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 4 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Dict, List, Optional
+
+import ulid
+
+PYSTAC_VERSION = "1.0.0"
```
paulfouquet, line 5: Could be a good idea to move this constant to a common Python file that could be use by `item.py` too.

paulfouquet (approved):

amfage (approved):

PR #166, opened 2022-10-04
refactor: single location for gdalinfo, refactor standardise_validate.py TDE-483

MDavidson17:
1

Code still runs as usual, should see no changes needed in topo workflows other than updating the version number. 

refactored to have gdalinfo in one location some changes have been done (splitting functions) in preparation to add stac item creation to standardise_validate.py

to run:
`docker run -v ${HOME}/.aws/credentials:/root/.aws/credentials:ro -e AWS_PROFILE=**profile** **container**:latest python standardise_validate.py --preset "webp" --source "[\"**s3://lpath/to.tif**\", \"**s3://path/to.tif**\"]"`

### DIFF
--- a/scripts/files/file_check.py
+++ b/scripts/files/file_check.py
@@ -89,8 +89,8 @@ class FileCheck:
     def validate(self, gdalinfo_result: Dict[Any, Any]) -> None:
         if self.is_valid():
             self.check_no_data(gdalinfo_result)
-            self.check_band_count(gdalinfo_result)
-            self.check_color_interpretation(gdalinfo_result)
+            # self.check_band_count(gdalinfo_result)
+            # self.check_color_interpretation(gdalinfo_result)
             gdalsrsinfo_tif_command = ["gdalsrsinfo", "-o", "wkt"]
             try:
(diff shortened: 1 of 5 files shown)


scripts/files/file_check.py
```diff
@@ -0,0 +1,99 @@
...
+        """Add an error if gdalsrsinfo and gdalsrsinfo_tif values are different.
+
+        Args:
+            gdalsrsinfo (str): Value returned by gdalsrsinfo as a string.
+            gdalsrsinfo_tif (str): Value returned by gdalsrsinfo for the tif as a string.
+        """
+        if gdalsrsinfo_tif != self.global_srs:
+            self.add_error(error_type="srs", error_message="different srs")
+
+    def check_color_interpretation(self, gdalinfo: Dict[str, Any]) -> None:
```
amfage, line 63: I think this will fail if the preset is `gray_webp` (the band checks were commented out before as we weren't sure they added any value, and they would need to be rewritten for grayscale imagery).

MDavidson17: Ah yes, I remember that, I can recomment them back out, until we work out what to do with them?

amfage (approved):
:+1:

PR #195, opened 2022-10-31
feat: rename files TDE-497

paulfouquet:
## Description
This PR implements the ability of renaming a file in both local and `s3` file system. The functionality is renaming a path to another, not only the file name. For `s3` the files have to be on the same bucket.

_**EDIT:** We don't need that at the moment. Keep this as a Draft for a bit._

### DIFF
--- a/scripts/files/fs_local.py
+++ b/scripts/files/fs_local.py
@@ -1,5 +1,7 @@
 import os
 
+from linz_logger import get_log
+
 
 def write(destination: str, source: bytes) -> None:
     """Write the source to the local destination file.
@@ -32,4 +34,21 @@ def rename(path: str, new_path: str) -> None:
         path (str): original path
         new_path (str): path to be renamed with
     """
+    if not os.path.exists(path):
+        get_log().error(
+            "rename_local_not_exists",
+            path=path,
+            destination=new_path,
+            error="The file to rename does not exist.",
+        )
+        raise Exception(f"{path} does not exist.")
+    if os.path.exists(new_path):
+        get_log().error(
+            "rename_local_already_exists",
+            path=path,
+            destination=new_path,
+            error="The destination file already exists.",
+        )
+        raise Exception(f"{new_path} already exists.")
+
     os.rename(src=path, dst=new_path)
--- a/scripts/files/fs_local.py
+++ b/scripts/files/fs_local.py
@@ -1,3 +1,6 @@
+import os
+
+
 def write(destination: str, source: bytes) -> None:
     """Write the source to the local destination file.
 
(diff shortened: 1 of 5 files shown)


scripts/files/fs_local.py
```diff
@@ -20,3 +25,30 @@ def read(path: str) -> bytes:
...
+
+
+def rename(path: str, new_path: str) -> None:
+    """Rename a file.
+
+    Args:
+        path (str): original path
+        new_path (str): path to be renamed with
+    """
+    if not os.path.exists(path):
```
blacha, line 37: checking if a file exists before doing something with it is somewhat of a antipattern.

between the time you check and the time you access it, the file could have been moved.

you are best to handle it in the actual action `os.rename` (Im assuming it throws if the file doesnt exist?

does it also throw if the target does exist?

scripts/files/fs_s3.py
```diff
@@ -67,6 +67,96 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
...
+                destination=new_path,
+                error="A file for the destination path already exists.",
+            )
+            raise Exception(f"{new_path} already exists. The file can't be renamed with this path.")
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                # OK: the destination file does not already exists
+                dst_s3_object.copy({"Bucket": src_s3_path.bucket, "Key": src_s3_path.key})
+                # delete the source
+                src_s3_object = s3.Object(src_s3_path.bucket, src_s3_path.key)
```
blacha, line 133: why is the source object recreated here? 

is it not the same object as https://github.com/linz/topo-imagery/pull/195/files#diff-3874c4d626707da994208cc5bced4772a5d0506c9b9912a63ff645f0b05450e7R104

scripts/files/fs_s3.py
```diff
@@ -67,6 +67,96 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
...
+            path=path,
+            destination=new_path,
+            error="The source and destination path are not in the same bucket",
+        )
+        raise Exception(f"Files {path} and {new_path} are not on the same S3 bucket.")
+
+    s3 = boto3.resource("s3")
+    try:
+        if needs_credentials:
+            s3 = get_session(path).client("s3")
```
blacha, line 101: this is only going to work if the source and targets are the same bucket?

scripts/files/fs_s3.py
paulfouquet: I forgot to remove that as I did push a commit later to check if the source object exists

scripts/files/fs_s3.py
paulfouquet: Yes

scripts/files/fs_local.py
paulfouquet: Looks like it does manage those: https://docs.python.org/3/library/os.html#os.rename

paulfouquet: > does it also throw if the target does exist?
I was looking at this now and

It does override the file as per
> If both are files, dst will be replaced silently if the user has permission.

I'm not too sure of what behaviour we want to have for the general functionality? 
I think for our context of renaming the tiff files as we don't want any duplicate we should probably do a check of duplicate files before the process of validate the tiff anyway?

PR #209, opened 2022-11-08
feat: rename tiff file if not aligned with its tile name TDE-497

paulfouquet:
## Description
Some tiff files does not respect the naming convention. This functionality check the correct name against the tile name link to the tiff and rename the tiff file if the name is not correct.
The associated workflow in Argo has to be updated once this is merged.
`non_visual_qa.py` has been deleted so associated workflow has to be deleted when this PR is merged.


scripts/files/file_check.py
```diff
@@ -81,19 +120,36 @@ def check_color_interpretation(self, gdalinfo: Dict[str, Any]) -> None:
...
                 custom_fields={"missing": f"{', '.join(missing_bands)}"},
             )
 
-    def validate(self, gdalinfo_result: Dict[Any, Any]) -> None:
-        if self.is_valid():
-            self.check_no_data(gdalinfo_result)
-            self.check_band_count(gdalinfo_result)
-            self.check_color_interpretation(gdalinfo_result)
+    def check_tile_and_rename(self, gdalinfo: Dict[Any, Any]) -> None:
+        origin = Point(gdalinfo["cornerCoordinates"]["upperLeft"][0], gdalinfo["cornerCoordinates"]["upperLeft"][1])
```
paulfouquet, line 129: Check if the origin is valid

amfage: Two examples of files with origin "drift":
`s3://linz-imagery-upload/Tasman Nelson Floods 2022 LiDAR & Orthos/renamed/top-of-the-south-flood/tifs/BP27_1000_4817.tif`
`s3://linz-imagery-upload/Invercargill2022_Pgrm3016/renamed/InvercargillUrban2022/tifs-RGB/CH10_500_001052.tif`

MDavidson17 (approved):

amfage (approved):

scripts/files/file_check.py
amfage: We need to work on the feature to correct the origin drift sooner rather than later otherwise files with origin drift will be renamed e.g. `file name "BP27_1000_4817.tiff"` with `origin = Point(1643679.999967818148434, 5444159.999954843893647)` will be renamed to `BP27_1000_4816.tiff`

amfage (approved):
Just one note inline about origin drift.

scripts/files/file_check.py
paulfouquet: We could create an extra check using a regex to check if the file name match the correct pattern ({tile_name}_{scale}_{originID} - like what is done in `argo-task` https://github.com/linz/argo-tasks/blob/master/src/commands/tileset-validate/tileset.validate.ts) AND <> than generated name, and logging or raising a specific issue for those cases?
@dwsilk

amfage (dismissed):

amfage (approved):

PR #210, opened 2022-11-09
feat: log the gdal version when run standardising TDE-559

paulfouquet:
## Description
Assuming that our container has the same GDAL version for the different command we run, logging the result of `gdalinfo --version` to get the GDAL version we use in our logs.

### DIFF
--- a/scripts/gdal/gdal_helper.py
+++ b/scripts/gdal/gdal_helper.py
@@ -49,11 +49,10 @@ def get_gdal_version() -> str:
     gdalinfo_version = ["gdalinfo", "--version"]
     try:
         proc = subprocess.run(gdalinfo_version, env=gdal_env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        return proc.stdout.decode().strip()
     except subprocess.CalledProcessError as cpe:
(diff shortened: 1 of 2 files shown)


scripts/gdal/gdal_helper.py
```diff
@@ -36,6 +36,26 @@ def command_to_string(command: List[str]) -> str:
...
+    """
+    gdal_env = os.environ.copy()
+    gdalinfo_version = ["gdalinfo", "--version"]
+    try:
+        proc = subprocess.run(gdalinfo_version, env=gdal_env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError as cpe:
+        get_log().error("get_gdal_version_failed", command=command_to_string(gdalinfo_version), error=str(cpe.stderr, "utf-8"))
+        raise GDALExecutionException(f"GDAL {str(cpe.stderr, 'utf-8')}") from cpe
+
+    return proc.stdout.decode().strip()
```
blacha, line 56: this is somewhat weird in a blocked scope lang this would fail as `proc` is not defined in this scope as it was only defined in the  `try` scope.

It would be best to put this into the `try`: scope

paulfouquet: Oh that's a mistake

blacha (approved):

PR #232, opened 2022-11-29
feat: log gdalinfo non visual qa error tde 563

paulfouquet:
Logging gdalinfo when there is error in the Non Visual QA so TDP can easily access to it.


paulfouquet:
sorry about that my branch is screwed up

PR #233, opened 2022-11-29
feat: log gdalinfo non visual qa errors TDE-563

paulfouquet:
Logging gdalinfo when there is error in the Non Visual QA so TDP can easily access to it.

### DIFF
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -10,6 +10,7 @@ from scripts.files.file_check import FileCheck
 from scripts.files.files_helper import is_tiff
 from scripts.files.fs import write
 from scripts.gdal.gdal_helper import get_srs
+from scripts.gdal.gdalinfo import format_wkt
 from scripts.standardising import start_standardising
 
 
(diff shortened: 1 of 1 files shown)


scripts/standardise_validate.py
```diff
@@ -50,14 +50,19 @@ def main() -> None:
         # Validate the file
         file_check = FileCheck(file, scale, srs)
         if not file_check.validate():
-            get_log().info("non_visual_qa_errors", file=file_check.path, errors=file_check.errors)
+            # Format gdalinfo for logging
+            formatted_gdalinfo = file_check.get_gdalinfo()
+            if formatted_gdalinfo:
+                formatted_gdalinfo["coordinateSystem"]["wkt"] = format_wkt(formatted_gdalinfo["coordinateSystem"]["wkt"])
+                formatted_gdalinfo["stac"]["proj:wkt2"] = format_wkt(formatted_gdalinfo["stac"]["proj:wkt2"])
+            get_log().info("non_visual_qa_errors", file=file_check.path, errors=file_check.errors, gdalinfo=formatted_gdalinfo)
```
paulfouquet, line 58: Elastic Search does not like that a (some?) field name is empty:
`{"errors":{"bands":[{"band":1,"block":[512,512],"colorInterpretation":"Red","maximum":255,"mean":135.328,"metadata":{"":{ [...]`

paulfouquet: Not a good idea to log the full gdalinfo json anyway

PR #239, opened 2022-12-01
feat: log standardised image gdal virtual path when non visual qa error TDE-563

paulfouquet:
## Description
In order to facilitate TDP team to investigate on Non Visual QA error, providing the GDAL virtual path of the standardized image located on the Argo artifact bucket will allow them to execute `gdalinfo` to that image.

> **_NOTE:_**  Another PR will be done to log the original file GDAL virtual path in the same log line so they can compare between original / standardized as well.

### DIFF
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -50,7 +50,13 @@ def main() -> None:
         # Validate the file
         file_check = FileCheck(file, scale, srs)
         if not file_check.validate():
-            get_log().info("non_visual_qa_errors", file=file_check.path, errors=file_check.errors)
+            vfs_path = ""
+            env_argo_template = os.environ.get("ARGO_TEMPLATE")
+            if env_argo_template:
+                argo_template = json.loads(env_argo_template)
+                s3_information = argo_template["archiveLocation"]["s3"]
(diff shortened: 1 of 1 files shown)


scripts/standardise_validate.py
```diff
@@ -47,7 +47,7 @@ def main() -> None:
             continue
 
         # Validate the file
-        file_check = FileCheck(file, scale, srs)
+        file_check = FileCheck(file, int(arguments.scale), srs)
         if not file_check.validate():
             vfs_path = ""
             env_argo_template = os.environ.get("ARGO_TEMPLATE")
```
amfage, line 53: There is an [`is_argo()` function](https://github.com/linz/topo-imagery/blob/47e966a675f40a9da578fbbabb612f9dc0e672f1/scripts/cli/cli_helper.py#L38).
Do we need to log this if it's run in a container outside of Argo?

paulfouquet: We mostly run it outside Argo for testing, so I did not want to duplicate the log line in that case. Instead it will log an empty path

paulfouquet: `is_argo()` returns a boolean, using it will call `os.environ.get()` twice in this case

amfage: Can you use it on line 54 instead? And remove this check?

amfage (approved):

PR #240, opened 2022-12-01
feat: apply a cutline if supplied

blacha:
- refactor: add basic types to gdalinfo response
- feat: detect the band count in imagery
- wip: attempt to apply a cutline

### DIFF
--- a/scripts/gdal/gdal_helper.py
+++ b/scripts/gdal/gdal_helper.py
@@ -57,14 +57,14 @@ def get_gdal_version() -> str:
 
 def run_gdal(
     command: List[str],
-    input_file: Optional[str] = None,
+    input_file: str,
     output_file: Optional[str] = None,
 ) -> "subprocess.CompletedProcess[bytes]":
     """Run the GDAL command. The permissions to access to the input file are applied to the gdal environment.
 
     Args:
         command (List[str]): each arguments of the GDAL command.
-        input_file (str, optional): the input file path.
+        input_file str: the input file path.
         output_file (str, optional): the output file path.
 
     Raises:
@@ -102,7 +102,7 @@ def run_gdal(
     if proc.stderr:
         get_log().warning("run_gdal_stderr", command=command_to_string(temp_command), stderr=proc.stderr.decode())
 
-    get_log().debug("run_gdal_succeeded", command=command_to_string(temp_command), stdout=proc.stdout.decode())
+    get_log().trace("run_gdal_succeeded", command=command_to_string(temp_command), stdout=proc.stdout.decode())
 
     return proc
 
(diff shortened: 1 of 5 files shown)


scripts/gdal/gdal_helper.py
```diff
@@ -102,7 +102,7 @@ def run_gdal(
     if proc.stderr:
         get_log().warning("run_gdal_stderr", command=command_to_string(temp_command), stderr=proc.stderr.decode())
 
-    get_log().debug("run_gdal_succeeded", command=command_to_string(temp_command), stdout=proc.stdout.decode())
+    get_log().trace("run_gdal_succeeded", command=command_to_string(temp_command), stdout=proc.stdout.decode())
```
blacha, line 105: this was spamming my console with the full gdal_info output, so I lowered the level, happy to raise it again if we really want it?

scripts/gdal/gdal_preset.py
```diff
@@ -1,122 +1,141 @@
...
+COMPRESS_LZW = [
+    # Compress as LZW
     "-co",
-    "overview_resampling=lanczos",
-    "-co",
-    "blocksize=512",
-    "-co",
-    "overview_quality=90",
+    "compress=lzw",
+    # Predictor two reduces file size
```
blacha, line 51: TODO find a better description as to why predictor two is better.

dwsilk: `PREDICTOR=2` stores the difference to the previous value rather than the actual value. `PREDICTOR=1` stores the actual value and `PREDICTOR=3` rearranges bits in float representations to optimise compression for floating point numbers and wouldn't have any effect for integers.

scripts/gdal/gdal_preset.py
```diff
@@ -1,122 +1,141 @@
...
+    alpha_band = find_band(bands, 'Alpha')
+    alpha_band_info: List[str] = []
+    if alpha_band:
+        alpha_band_info.extend(['-b',  str(alpha_band['band'])])
+
+    # Grey scale imagery, set R,G and B to just the grey_band
+    grey_band = find_band(bands, 'Gray')
+    if grey_band:
+        grand_band_index = str(grey_band['band'])
+        return ["-b", grand_band_index, "-b", grand_band_index, "-b", grand_band_index] + alpha_band_info
```
dwsilk, line 118: `grand_band_index`? `grey_band_index`?

dwsilk (changes requested):

scripts/gdal/gdal_helper.py
paulfouquet: That's a good call. The output formatting is also very bad when calling gdalinfo with json output

paulfouquet (dismissed):

paulfouquet (approved):

PR #258, opened 2022-12-12
feat: optimise cutline handling by only using the cutline if required

blacha:


scripts/gdal/gdal_cutline.py
```diff
@@ -0,0 +1,79 @@
...
+    raster_area = tiff_geom.GetArea();
+
+    intersection_area = intersection_area / raster_area 
+
+    # 99.99% of the tiff is covered by the cutline, so we likely dont need the cutline
+    if intersection_area >= 0.9999:
+        return None
+
+    # Not asked to optimize the cutline
+    if not optimize:
```
blacha, line 60: I have left this disabled for now, as it needs more testing and just by ignoring the cutline for most tiles should significantly speed up the process

blacha:
This logic can be removed once https://github.com/OSGeo/gdal/issues/6905 is released.

scripts/gdal/gdal_cutline.py
```diff
@@ -0,0 +1,101 @@
+from typing import Optional
+
+# osgeo is embbed in the Docker image
+from osgeo import gdal  # pylint: disable=import-error
+from osgeo import Geometry, Layer, ogr  # pylint: disable=import-error
```
paulfouquet, line 5: You could import everything from `osgeo` in the same line

paulfouquet (approved):

PR #285, opened 2023-01-06
build: devcontainer TDE-598

paulfouquet:
## Description
Add a `devcontainter` to this repo in order to facilitate development in a Python with GDAL and Poetry dependencies environment.

### DIFF
--- a/.devcontainer/post_start.sh
+++ b/.devcontainer/post_start.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+git config --global --add safe.directory $1
+poetry run pre-commit install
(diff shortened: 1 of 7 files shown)


.devcontainer/post_start.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+git config --global --add safe.directory $1
```
paulfouquet, line 3: https://github.com/microsoft/vscode-remote-release/issues/7628

paulfouquet (commented):
Converted to draft as there is an issue with `pre-commit install` during the very first initialization of the container.
To reproduce, need to clean Docker cache `docker builder prune`

paulfouquet:
Using `devcontainer` made my VSCode crashing few times. I've opened https://github.com/microsoft/vscode/issues/170999

.devcontainer/post_start.sh
paulfouquet: The fix in devcontainer pre-release mentioned in https://github.com/microsoft/vscode-remote-release/issues/7628#issuecomment-1406157351 did not work for me

PR #299, opened 2023-01-12
ci: publish container with all version tags

paulfouquet:
This has been tested here https://github.com/paulfouquet/topo-imagery-test

### DIFF
--- a/.github/scripts/publish_container.sh
+++ b/.github/scripts/publish_container.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+GIT_VERSION=$(git describe --tags --always --match 'v*')
+XYZ_VERSION=${GIT_VERSION%-*-*}
+XY_VERSION=${XYZ_VERSION%.*}
+X_VERSION=${XY_VERSION%.*}
+
+for version in "latest" ${X_VERSION} ${XY_VERSION} ${XYZ_VERSION} ${GIT_VERSION}; do
+  docker tag topo-imagery ghcr.io/linz/topo-imagery:$version
+  docker push ghcr.io/linz/topo-imagery:$version
+done
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -21,9 +21,5 @@ jobs:
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
 
       - name: Publish Containers
-        run: |
-          GIT_VERSION=$(git describe --tags --always --match 'v*')
-          docker tag topo-imagery ghcr.io/linz/topo-imagery:${GIT_VERSION}
-          docker push ghcr.io/linz/topo-imagery:${GIT_VERSION}
-          docker tag topo-imagery ghcr.io/linz/topo-imagery:latest
-          docker push ghcr.io/linz/topo-imagery:latest
+        run: ./.github/scripts/publish_container.sh
+        shell: bash


.github/scripts/publish_container.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
blacha, line 1: I would leave this inside the containers.yaml

.github/scripts/publish_container.sh
```diff
@@ -0,0 +1,11 @@
...
+
+GIT_VERSION=$(git describe --tags --always --match 'v*')
+XYZ_VERSION=${GIT_VERSION%-*-*}
+XY_VERSION=${XYZ_VERSION%.*}
+X_VERSION=${XY_VERSION%.*}
+
+for version in "latest" ${X_VERSION} ${XY_VERSION} ${XYZ_VERSION} ${GIT_VERSION}; do
```
blacha, line 8: we only want `latest` and `GIT_VERSION` on master

we want the `X` and `XY` and `XYZ` on release

.github/scripts/publish_container.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+GIT_VERSION=$(git describe --tags --always --match 'v*')
+XYZ_VERSION=${GIT_VERSION%-*-*}
```
blacha, line 4: what happens when GIT_VERSION is `v1.0.0` vs `v1.0.0-12-gdeff9a1`

paulfouquet: We will create the tag `v1.0.0` twice, don't know how it's handled

paulfouquet: I've made a test and it does not seem to tag the container twice, event if we run the command twice:
image

If I find a better way to handle the `GIT_VERSION` I'll change that but I think for now that's enough?

.github/workflows/containers.yml
```diff
@@ -22,8 +22,18 @@ jobs:
 
       - name: Publish Containers
         run: |
+          versions=("latest")
           GIT_VERSION=$(git describe --tags --always --match 'v*')
-          docker tag topo-imagery ghcr.io/linz/topo-imagery:${GIT_VERSION}
-          docker push ghcr.io/linz/topo-imagery:${GIT_VERSION}
-          docker tag topo-imagery ghcr.io/linz/topo-imagery:latest
-          docker push ghcr.io/linz/topo-imagery:latest
+          if [ ${{(github.ref == 'refs/heads/master' && contains(github.event.head_commit.message, 'release:'))}} == true ]; then
```
paulfouquet, line 27: Have to use contains as when the `release-please` PR gets merged, the commit message not only contains the `release: ...` commit:
```
commit 68f58f69bbaec5a5895a1ef3572949ec516ac39a (HEAD -> master, tag: v0.4.0, origin/master, origin/HEAD)
Merge: 94fb2a7 1ca37f3
Author: paulfouquet <86932794+paulfouquet@users.noreply.github.com>
Date:   Mon Jan 16 10:54:41 2023 +1300

    Merge pull request #29 from paulfouquet/release-please--branches--master
    
    release: 0.4.0
```

blacha: your commit  here looks to be a Merge commit, we do not allow merge commits in this repo.

When merging it should be merged as a Squash+Rebase, here is an example commit with a squash/rebase

https://github.com/linz/style-js/commit/169ad32dd68de237433e4079bca0ffe29f8f4569

so `startsWith` is what we want.

paulfouquet: I haven't do my test in a exact same environment that's why...

.github/workflows/release-please.yml
```diff
@@ -15,10 +15,4 @@ jobs:
         with:
           release-type: python
           token: ${{ secrets.GITHUB_TOKEN }}
-
-  publish-release:
```
paulfouquet, line 19: If this was intended to publish the tags related to a release, i.e. `vX`, `vX.Y` and `vX.Y.Z`, that would be a duplicate as it is handled directly in `container.yml` without needing another step to trigger it.
Another way would be to duplicate the bash script in `container.yml` in here but handling the `X`, `XY` and `XYZ` tags.

.prettierignore
```diff
@@ -0,0 +1 @@
+CHANGELOG.md
```
paulfouquet, line 1: This is to avoid `prettier` to complain about `*` VS `-` in the changelog generated by `release-please`

.github/workflows/containers.yml
```diff
@@ -18,12 +18,14 @@ jobs:
           docker build . --tag topo-imagery --label "github_run_id=${GITHUB_RUN_ID}"
 
       - name: Log in to registry
+        if: ${{(github.ref == 'refs/heads/master') && !(startsWith(github.event.head_commit.message, 'release:'))}}
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
 
       - name: Publish Containers
+        if: ${{(github.ref == 'refs/heads/master') && !(startsWith(github.event.head_commit.message, 'release:'))}}
```
paulfouquet, line 25: This is to avoid a tag being created at this stage when a release is published.

blacha (approved):

PR #333, opened 2023-02-02
feat: standardising 16bits TIFF TDE-628

paulfouquet:
We recently had a dataset with 16 bits TIFF (band.type = UInt16). gdal_translate with webp compression requires a 8 bits TIFF. The solution proposed here is to determine the number of bits of the first band of the TIFF, assuming all the bands have the same type and in the case of a UInt16 type, apply the transformation from 16 to 8 bits.

Test output: https://basemaps.linz.govt.nz/?config=MwKXXnaFtYQhoQ4PfLzAcwBJmdCMnAyh4mwgDAi863PxdHbPqRh8Gdmc27zA9GKTJQQKf2XPqgBb4x8e7z8rDkoYVAwJBb8bFztxBJuZhV4x36bbTLka64zzphgmwTP3eXj7T5AKxwai7ANWEfxghFAdTDPKdD4pMPTEpdRi7LD2yeqbSyrcJDxMzjMqHgtgbA5rroncmGMTXE9ay&i=test-queenstown-lake-0.1m-urban-aerial-photos-2022&tileMatrix=NZTM2000Quad&debug#@-44.679146,169.193535,z9


scripts/standardising.py
```diff
@@ -100,8 +101,14 @@ def standardising(file: str, preset: str, cutline: Optional[str]) -> FileTiff:
...
 
-        command = get_gdal_command(preset)
-        command.extend(get_gdal_band_offset(input_file))
+            # gdalinfo to get band offset and band type
+            info = gdal_info(file, False)
+            convert_to_byte = False
+            if get_gdal_band_type(input_file, info) == "UInt16":
+                convert_to_byte = True
+
+            command = get_gdal_command(preset, convert_to_byte)
```
blacha, line 110: I get a little nervous passing booleans into functions, could this be a keyword argument eg `covert_to_byte=convert_to_byte` 

Makes it more clear that the boolean is very specific to this conversion.

also,  there could be a lot of "convert_x_to_byte" do we want to be a little more specific that this only works for uint16s ?

paulfouquet: I had this thought, but as there is currently only one conversion and I could not predict if there will be other, I preferred not being specific yet.
I'm happy to change that if it's too confusing

paulfouquet: > I get a little nervous passing booleans into functions, could this be a keyword argument eg covert_to_byte=convert_to_byte

Passing the input band type to `get_gdal_command(preset, band_type="UInt16")` and dealing with it in `get_gdal_command` would suit?

blacha: I like your second option!

`convert_from` or `input_band_type` both sound like good options!

amfage (approved):

MDavidson17 (approved):

PR #363, opened 2023-02-19
fix: be specific about nodata being None

MDavidson17:


scripts/files/file_tiff.py
```diff
@@ -101,7 +101,8 @@ def check_no_data(self, gdalinfo: GdalInfo) -> None:
     def is_no_data(self, gdalinfo: GdalInfo) -> bool:
         """return True if bands have a "noDataValue" and it is set to 255."""
```
amfage, line 102: Should the comment be updated?

scripts/files/file_tiff.py
```diff
@@ -101,7 +101,8 @@ def check_no_data(self, gdalinfo: GdalInfo) -> None:
...
         bands = gdalinfo["bands"]
-        if "noDataValue" in bands[0] and bands[0]["noDataValue"]:
+        # 0 in noDataValue can return false unless specific here about None
+        if "noDataValue" in bands[0] and not bands[0]["noDataValue"] is None:
```
paulfouquet, line 105: To me it's more readable using `bands[0]["noDataValue"] is not None` 
I also think that's what is recommended https://peps.python.org/pep-0008/#programming-recommendations

paulfouquet (approved):

amfage (approved):

PR #375, opened 2023-02-28
feat: take source and target epsg, if not the same transform TDE-699

MDavidson17:


paulfouquet (approved):

scripts/gdal/gdal_preset.py
```diff
@@ -76,12 +73,12 @@
...
 
 
-def get_gdal_command(preset: str, convert_from: Optional[str] = None) -> List[str]:
+def get_gdal_command(preset: str, source_epsg: str, convert_from: Optional[str] = None) -> List[str]:
     get_log().info("gdal_preset", preset=preset)
     gdal_command: List[str] = ["gdal_translate"]
 
     gdal_command.extend(BASE_COG)
-    gdal_command.extend(NZTM_SOURCE)
+    gdal_command.extend(["-a_srs", f"EPSG:{source_epsg}"])
```
paulfouquet, line 81: Can you put back the comment `# Force the source projection to source_epsg`

amfage (approved):

scripts/gdal/gdal_preset.py
```diff
@@ -132,3 +130,10 @@ def get_alpha_command() -> List[str]:
         # Ensure the target has a alpha channel
         "-dstalpha",
     ]
+
+
+def get_transform_srs_command(target_epsg: str) -> List[str]:
+    """
+    Get a "Gdalwarp" command to transform the srs
+    """
+    return ["gdalwarp", "-t_srs", f"EPSG:{target_epsg}"]
```
blacha, line 139: What resampling method does `gdalwarp` run by default? I believe we need to use `bilinear` (?)

MDavidson17: yes I was just thinking this, shall we match Rebecca's gist:
`gdalwarp -t_srs 'EPSG:2193' -r bilinear -tr 10 -10 -co compress=lzw ./output/overall_wgs84.tif ./output/overall_NZTM.tif`

blacha: this is also missing the rest of the arguments, like `ALL_CPUS` that I shared earlier.

blacha (approved):

PR #399, opened 2023-03-20
fix: use multithreading for gdalwarp when creating COGS TDE-696

blacha:


scripts/gdal/gdal_preset.py
```diff
@@ -18,6 +18,10 @@
     # Ensure all CPUs are used for gdal translate
     "-co",
     "num_threads=all_cpus",
+    # Until GDAL 3.7.x this needs to be set as well as num_threads.
```
paulfouquet, line 21: Do you have a link to the doc or github issue describing that?

paulfouquet (approved):

scripts/gdal/gdal_preset.py
paulfouquet, line 21:
```suggestion
    # Until GDAL 3.7.x this needs to be set as well as num_threads (https://github.com/OSGeo/gdal/issues/7478)
```

paulfouquet (approved):

PR #403, opened 2023-03-21
feat: import sidecar xml metadata to get cloud percentage into stac TDE-683

paulfouquet:
- take two new arguments `--metadata-file` and `--metadata-type`
- an `earthscanner` type has been added
- parse the cloud percentage to put it in the stac metadata with the `eo` extension


scripts/stac/imagery/item.py
```diff
@@ -46,3 +51,20 @@ def add_collection(self, collection_id: str) -> None:
...
     def add_link(self, rel: str, href: str = "./collection.json", file_type: str = "application/json") -> None:
         self.stac["links"].append({"rel": rel, "href": href, "type": file_type})
+
+    def add_stac_extension(self, link: str) -> None:
+        """add a stac extension link in the 'stac_extensions' list
+
+        Args:
+            link: to the extension json schema
+        """
+        self.stac["stac_extensions"].append(link)
```
blacha, line 61: could be good to check if the extension was already in the list before appending it again.

scripts/stac/imagery/import_metadata.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class MetadataType(str, Enum):
+    EARTHSCANNER = "earthscanner"
+
+
+def get_cloud_percent(xml_input: str) -> Optional[int]:
+    try:
+        input_metadata = xmltodict.parse(xml_input)
+        cloud_percent: int = int(input_metadata["MetaData"]["ProductInfo"]["CloudPercent"])
```
blacha, line 15: do we know if this is always a `int` ? could it be a `float` ?

blacha (dismissed):

scripts/standardise_validate.py
```diff
@@ -27,6 +27,8 @@ def main() -> None:
     parser.add_argument(
         "--end-datetime", dest="end_datetime", help="end datetime in format YYYY-MM-DD", type=valid_date, required=True
     )
+    parser.add_argument("--metadata-file", dest="metadata_file", help="Sidecar metadata file")
+    parser.add_argument("--metadata-type", dest="metadata_type", help="Sidecar metadata file")
```
blacha, line 31: might also want to expand on the help docs for this, could we say what are some valid options?

scripts/stac/imagery/import_metadata.py
paulfouquet: From the examples I've looked at, they were always integer. For example `JL1KF01C_PMSR3_20230308055539_200139323_102_0020_001_L3C_PSH_meta.xml` has a value of `14`. I could parse it as a float, but wonder if want to have a value like `14.00` in the stac metadata for this instance?

scripts/standardise_validate.py
paulfouquet: We haven't documented the CLI arguments on this repo. They are undirectly documented in https://github.com/linz/topo-workflows/blob/master/workflows/imagery/README.md where a change needs to be made to use these two new arguments. For the valid options, I think having them as an enum in the workflow will give direction to the user and avoid any typing/spelling issue to match the values in `MetadataType`.

paulfouquet:
Put as a Draft as blocked https://toitutewhenua.atlassian.net/browse/TDE-683?focusedCommentId=109540

PR #432, opened 2023-04-04
fix: fs_s3.read() exceptions are not caught properly

paulfouquet:
Looking into fixing the `fs_s3` unit tests, noticed that the exceptions were not caught properly.

### DIFF
--- a/scripts/files/fs_s3.py
+++ b/scripts/files/fs_s3.py
@@ -39,11 +39,14 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
     """Read a file on a AWS S3 bucket.
 
     Args:
-        path (str): The AWS S3 path to the file to read.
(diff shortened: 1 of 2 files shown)


amfage (approved):

scripts/files/fs_s3.py
```diff
@@ -56,17 +59,18 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
 
         s3_object = s3.Object(s3_path.bucket, key)
         file: bytes = s3_object.get()["Body"].read()
+    except s3.meta.client.exceptions.NoSuchBucket as nsb:
+        get_log().error("s3_bucket_not_found", path=path, error=f"The specified bucket does not seem to exist: {nsb}")
```
MDavidson17, line 63: nit: I see this is just copied but, could this be "The specified bucket does not exist"?

MDavidson17 (approved):

PR #441, opened 2023-04-12
feat: fs.exists() checks if path exists TDE-710

paulfouquet:
`fs.exists()` checks if a `s3` or local path points to an existing file/directory/object

### DIFF
--- a/scripts/files/fs_s3.py
+++ b/scripts/files/fs_s3.py
@@ -76,6 +76,46 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
     return file
 
 
+def exists(path: str, needs_credentials: bool = False) -> bool:
+    """Check if s3 Object exists
+
+    Args:
+        path: path to the s3 object/key
+        needs_credentials: if acces to object needs credentials. Defaults to False.
+
+    Raises:
+        ce: ClientError
+        nsb: NoSuchBucket
+
+    Returns:
+        True if the S3 Object exists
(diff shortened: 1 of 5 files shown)


scripts/files/fs_s3.py
```diff
@@ -76,6 +76,46 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
...
+    s3_path, key = parse_path(path)
+    s3 = boto3.resource("s3")
+
+    try:
+        if needs_credentials:
+            s3 = get_session(path).resource("s3")
+
+        bucket_name = bucket_name_from_path(s3_path)
+        bucket = s3.Bucket(bucket_name)
+        objects = bucket.objects.filter(Prefix=key)
```
blacha, line 102: generally you would do a `head` to see if a key exists, this is checking  for anything that starts with this path

so `exists s3://foo/bar.tiff`  would succeed if you did `exists s3://foo/bar.tiff.bak`

paulfouquet: Oh, that's right.
I'm going to fix that

paulfouquet: @blacha this is now checking if the `path` is a "directory" aka ends with `/` to determine wether it does a `head` or a `list`.

amfage (approved):

MDavidson17 (approved):

amfage (approved):

PR #444, opened 2023-04-12
feat!: resume standardise validate TDE-710

paulfouquet:
Allows `standardise_validate.py` to be resumed using the same `target` (output directory) after a failure in order to avoid re-process the entire `source`.

## Changes

- :warning:  Takes a `target` directory/`s3` path as argument where the output files (TIFFS and STAC metadata `.json`) will be saved
- verify that the `target` exists to enter in the "resume" mode
- if "resume" mode, both STAC and TIFF files existance are check: 
  - if STAC exists, the TIFF is considered to exists too and the source file will be ignored (not processed)
  - if STAC does not exists but TIFF exists, the TIFF will be validated (Non Visual QA) and its STAC metadata created

> **_NOTE:_**  Draft PR to modify the `standardising` Argo Workflow: https://github.com/linz/topo-workflows/pull/123

## Tests
Test Argo Workflows: https://argo.linzaccess.com/workflows/argo/test-imagery-standardising-resume-gt8wr?tab=workflow


scripts/standardise_validate.py
```diff
@@ -40,12 +85,31 @@ def main() -> None:
     if is_argo():
         concurrency = 4
 
+    # The target exists, consider this is a resumed workflow
+    is_resuming = exists(arguments.target)
```
paulfouquet, line 89: This might be not helpfull as if one task did create at least one file, thie target directory will exists so all new instance of the script will be in a "resume" mode

scripts/files/fs.py
```diff
@@ -28,3 +28,17 @@ def read(path: str) -> bytes:
         return fs_s3.read(path)
 
     return fs_local.read(path)
+
+
+def exists(path: str) -> bool:
+    """Check if path (file or directory) exists
+
+    Args:
+        path: A path to a directory
```
blacha, line 37: NIT: or file?

scripts/standardise_validate.py
blacha: I think this may be better handled in the `run_standardising` where each standardizing checks to see if the file exists before processing rather than trying to figure out what needs to process first.

scripts/standardise_validate.py
```diff
@@ -1,14 +1,58 @@
...
+            gdalinfo = gdal_info(path)
+            origin = Point(gdalinfo["cornerCoordinates"]["upperLeft"][0], gdalinfo["cornerCoordinates"]["upperLeft"][1])
+            tile_name = get_tile_name(origin, scale)
+            prefix = os.path.join(target, tile_name)
+            target_path_stac = prefix + ".json"
+            target_path_tiff = prefix + ".tiff"
+            # STAC item json file name
+            if exists(target_path_stac):
+                # STAC exists, consider the TIFF exists too
+                processed_tiffs.append(path)
```
blacha, line 45: it should validate that the tiff exists too. It's only complete if both are present.

MDavidson17 (approved):

amfage (approved):

PR #487, opened 2023-05-19
docs: add missing docstring

paulfouquet:
add some `docstring` to most of the functions in the entire codebase


scripts/files/file_tiff.py
```diff
@@ -36,15 +36,36 @@ def __init__(
...
+            path: the path to the standardised file
+        """
         self._path_standardised = path
 
     def get_gdalinfo(self) -> Optional[GdalInfo]:
+        """Get the `gdalinfo` output for the file.
+        Run gdalinfo if not already ran.
+
+        Returns:
+            the `gdalinfo` output
```
blacha, line 67: This returns `None` if there was a error?

scripts/files/file_tiff.py
```diff
@@ -36,15 +36,36 @@ def __init__(
         self._srs: Optional[bytes] = None
 
     def set_srs(self, srs: bytes) -> None:
+        """Set the srs.
```
blacha, line 39: would be good to explain what a `srs` is or link to it.

scripts/files/file_tiff.py
```diff
@@ -36,15 +36,36 @@ def __init__(
...
     def set_scale(self, scale: int) -> None:
+        """Set the scale.
+
+        Args:
+            scale: the scale as `int`
+        """
         self._scale = scale
 
     def set_path_standardised(self, path: str) -> None:
+        """Set the standardised file path.
```
blacha, line 55: an example of this would be great.

scripts/files/file_tiff.py
```diff
@@ -36,15 +36,36 @@ def __init__(
...
     def set_srs(self, srs: bytes) -> None:
+        """Set the srs.
+
+        Args:
+            srs: the srs in bytes
+        """
         self._srs = srs
 
     def set_scale(self, scale: int) -> None:
+        """Set the scale.
```
blacha, line 47: I think this could use further explanation, what is a `scale` what are normal values for it?

scripts/files/file_tiff.py
paulfouquet: I'm going fix that in another PR

amfage (approved):

PR #489, opened 2023-05-19
refactor: move create_stac.py script to /stac folder

paulfouquet:


amfage (approved):
Just a note that the path is referenced here as well https://github.com/linz/topo-workflows/blob/master/workflows/imagery/README.md#Standardising

paulfouquet:
> Just a note that the path is referenced here as well https://github.com/linz/topo-workflows/blob/master/workflows/imagery/README.md#Standardising

Good spotting! thanks https://github.com/linz/topo-workflows/pull/130

MDavidson17 (approved):

PR #522, opened 2023-06-26
feat: add dem_lerc  preset TDE-779

MDavidson17:
LERC GDAL Command: `gdal_translate -q -of COG -co blocksize=512 -co num_threads=all_cpus -co sparse_ok=true -co bigtiff=yes -a_srs EPSG:2193 -stats -co compress=lerc -co max_z_error=0.001`

Example Processed Tiff:
[CF15_1000_3748.zip](https://github.com/linz/topo-imagery/files/11888193/CF15_1000_3748.zip)

### DIFF
--- a/scripts/gdal/gdal_preset.py
+++ b/scripts/gdal/gdal_preset.py
@@ -29,6 +29,14 @@ BASE_COG = [
     "bigtiff=yes",
 ]
 
+DEM_LERC = [
+    "-stats",
+    "-co",
+    "compress=lerc",
+    "-co",
+    "max_z_error=0.001",
+]
+
 COMPRESS_LZW = [
     # Compress as LZW
     "-co",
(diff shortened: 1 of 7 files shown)


scripts/gdal/gdal_preset.py
```diff
@@ -29,6 +29,15 @@
     "bigtiff=yes",
 ]
 
+DEM_LERC = [
+    "-stats",
```
blacha, line 33: I think `-stats` should be on all standardization not just `lerc`

scripts/gdal/tests/gdal_preset_test.py
```diff
@@ -43,6 +43,26 @@ def test_preset_lzw() -> None:
     assert "EPSG:2193" in gdal_command
 
 
+def test_preset_dem_lerc() -> None:
+    gdal_command = get_gdal_command("dem_lerc", epsg="2193")
+    # Basic cog creation
+    assert "COG" in gdal_command
+    assert "blocksize=512" in gdal_command
+    assert "num_threads=all_cpus" in gdal_command
+    assert "bigtiff=yes" in gdal_command
```
blacha, line 52: I am wondering if we need `bigtiff=yes` do we ever plan on making tiffs bigger than 4GB? I would hope not. so maybe we should make this `no` and then GDAL will fail if we ever accidentally make a tiff too big ?

MDavidson17: @dwsilk, in a huddle we could not remember/find reason why we have `Bigtiff` always set to `yes`, do you have any thoughts/can recall?

rebclarke: Not something we used in our original process. The raster files we work with are always under 4GB.

rebclarke: Only circumstances I can think we'll need it is if we're working with datasets like adhoc satellite imagery procured for emergency response.

amfage (approved):

amfage (approved):

scripts/gdal/tests/gdal_preset_test.py
MDavidson17: So that we can better document the conversation and in order to complete this ticket, this conversation will be continued in [TDE-805](https://toitutewhenua.atlassian.net/browse/TDE-805).

blacha: @MDavidson17 perfect, lets go ahead and merge this then!

PR #531, opened 2023-07-03
feat: add --producer-list to allow multiple producers TDE-781

ajacombs:
Adds `--producer-list` argument to `collection_from_items.py` to allow passing multiple producers to the STAC metadata.

There is a slight change to how both `--licensor-list` and `--producer-list` are handled, which resolves what I think is an edge case that is currently not handled.

Currently, if `--licensor-list` is passed a value without `;`, and `--licensor` is not passed, the output STAC file will include `{ "name": null, "roles": ["licensor"] }`. This PR fixes that, so that it will just not be present in the output.

Alternatively, it could be included as a single un-split item by removing the `and ";" in arguments.licensor_list` checks.

*Edit:* Which now that I've looked at where it's used in the standardising workflow I see would never eventuate in practice, as `--licensor` gets a default value of "Unknown" so `--licensor` is never not passed a value as the script is used in the standardising workflow (though could be if used elsewhere).

### DIFF
--- a/scripts/collection_from_items.py
+++ b/scripts/collection_from_items.py
@@ -19,23 +19,34 @@ def main() -> None:
     parser.add_argument("--collection-id", dest="collection_id", help="Collection ID", required=True)
     parser.add_argument("--title", dest="title", help="Collection title", required=True)
     parser.add_argument("--description", dest="description", help="Collection description", required=True)
-    parser.add_argument("--producer", dest="producer", help="Imagery producer", required=True)
+    parser.add_argument("--producer", dest="producer", help="Imagery producer")
+    parser.add_argument("--producer-list", dest="producer_list", help="Semicolon delimited list of imagery producers")
     parser.add_argument("--licensor", dest="licensor", help="Imagery licensor")
-    parser.add_argument("--licensor-list", dest="licensor_list", help="Imagery licensor list")
+    parser.add_argument("--licensor-list", dest="licensor_list", help="Semicolon delimited list of imagery licensors")
     parser.add_argument(
         "--concurrency", dest="concurrency", help="The number of files to limit concurrent reads", required=True, type=int
     )
 
     arguments = parser.parse_args()
     uri = arguments.uri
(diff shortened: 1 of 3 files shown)


paulfouquet (dismissed):

scripts/collection_from_items.py
```diff
@@ -14,28 +14,48 @@
...
         "--concurrency", dest="concurrency", help="The number of files to limit concurrent reads", required=True, type=int
     )
 
     arguments = parser.parse_args()
     uri = arguments.uri
-    licensors: List[str] = parse_list(arguments.licensor_list)
-    if len(licensors) <= 1:
-        licensors = [arguments.licensor]
+
+    producers: List[str] = []
```
blacha, line 42: I wonder if turning this into a function would be nicer?

```
def parse_list_(item_list: Optional[str], item_single: Optional[str]) -> List[str]:
 output = []
 if item_list:
￼  output.extend(parse_list(item_list)
￼elif item_single:
￼  output.append(item_single)
return output


for producer in parse_list(args.producer_list, args.producer):
  providers.append({ "name":... })
for licensor in parse_list(args.licensor_list, args.licensor):
  providers.append({ "name":... })
```

I also wonder if it would just be easier to allow `licensor` to be a `;` separated list

```
--licensor "Something;Something Else"
--licensor "Something"
--licensor '["Something","Something Else"]'
```

Rather than having two parameters? I guess the question would be how would that work in Argo?

paulfouquet: When I did the `producers` I ran into an issue with Argo not being able to wrap the list into `[ ... ]`

ajacombs: Have refactored argument parsing into a helper function.

Changing to single argument would involve adding conditional logic to Argo workflow template and/or not having a dropdown in the Argo GUI, so would be best done via another PR.

blacha (approved):

blacha (approved):

paulfouquet (approved):

PR #537, opened 2023-07-05
fix: AREA_OR_POINT should be forced to AREA for DEMs TDE-787

paulfouquet:


scripts/gdal/gdal_preset.py
```diff
@@ -36,6 +36,8 @@
     "-co",
     # Set Max Z Error to 1mm
     "max_z_error=0.001",
+    "-mo",
```
blacha, line 39:
```suggestion
    # Force all DEMS to AREA to be consistent 
    # input tiffs vary between AREA or POINT
    "-mo",
```

scripts/gdal/gdal_preset.py
```diff
@@ -36,6 +36,10 @@
     "-co",
     # Set Max Z Error to 1mm
     "max_z_error=0.001",
+    # Force all DEMS to AREA to be consistent
+    # input tiffs vary between AREA or POINT
+    "-mo",
+    "AREA_OR_POINT=AREA",
```
amfage, line 42: Complete NIT but the convention seems to be `AREA_OR_POINT=Area` (or `AREA_OR_POINT=Point`).

amfage (approved):

PR #310, opened 2023-01-24
feat: use -expand option for paletted tiffs

ajacombs:
These changes handle TIFFs with a band of `Palette` colour interpretation type, using the [`-expand`](https://gdal.org/programs/gdal_translate.html#cmdoption-gdal_translate-expand) argument to `gdal_translate`. This takes the contents of the colour palette and applies it to the rgb/rgba bands in the output file.

This resolves the following error encountered when standardising historical imagery survey SN9911:

> `ERROR 6: GTIFFBuildOverviews() doesn't support building overviews of multiple colormapped bands.`

which used the band arguements `-b 1 -b 1 -b 1`.

It seems plausible that there could be malformed paletted TIFFs which would not be able to be standardised using this change, but it resolves the direct issue for SN9911.

PR #547, opened 2023-07-12
feat: handle standardising paletted tiffs

ajacombs:
These changes handle TIFFs with a band of Palette colour interpretation type, using the [-expand](https://gdal.org/programs/gdal_translate.html#cmdoption-gdal_translate-expand) argument to gdal_translate. This takes the contents of the colour palette and applies it to the rgb/rgba bands in the output file.

This resolves the following error encountered when standardising historical imagery survey SN9911:

    ERROR 6: GTIFFBuildOverviews() doesn't support building overviews of multiple colormapped bands.

which used the band arguements -b 1 -b 1 -b 1.

This PR supersedes https://github.com/linz/topo-imagery/pull/310 which was proposed when previously processing historical imagery. I've made a new PR mainly to simplify having to rebase 6 months worth of changes from master onto that branch. 

Changes from the earlier PR are that I have now removed the raising of an exception and the handling to catch it, which were the aspects still under discussion preventing it being merged.

### DIFF
--- a/scripts/gdal/gdal_bands.py
+++ b/scripts/gdal/gdal_bands.py
@@ -36,19 +36,31 @@ def get_gdal_band_offset(file: str, info: Optional[GdalInfo] = None, preset: Opt
 
     bands = info["bands"]
 
-    alpha_band = find_band(bands, "Alpha")
-    alpha_band_info: List[str] = []
-    if alpha_band:
-        alpha_band_info.extend(["-b", str(alpha_band["band"])])
+    band_alpha_arg: List[str] = []
+    if band_alpha := find_band(bands, "Alpha"):
+        band_alpha_arg.extend(["-b", str(band_alpha["band"])])
 
-    grey_band = find_band(bands, "Gray")
-    if grey_band:
-        grey_band_index = str(grey_band["band"])
+    if band_grey := find_band(bands, "Gray"):
+        band_grey_index = str(band_grey["band"])
         if preset == "dem_lerc":
             # return single band if DEM/DSM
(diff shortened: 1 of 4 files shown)


scripts/gdal/gdal_bands.py
```diff
@@ -36,19 +37,31 @@ def get_gdal_band_offset(file: str, info: Optional[GdalInfo] = None, preset: Opt
...
+        return ["-b", band_grey_index, "-b", band_grey_index, "-b", band_grey_index] + band_alpha_arg
+
+    if band_palette := find_band(bands, "Palette"):
+        if colour_table := band_palette["colorTable"]:
+            palette_channels = len(colour_table["entries"][0])
+            if palette_channels == 4:
+                return ["-expand", "rgba"]
+            if palette_channels == 3:
+                return ["-expand", "rgb"]
+            get_log().error(
```
blacha, line 59: should this raise a exception?

scripts/gdal/gdalinfo.py
```diff
@@ -8,6 +8,12 @@
 from scripts.tile.tile_index import Point
 
 
+class GdalInfoBandColorTable(TypedDict):
+    palette: str
```
paulfouquet, line 12: Could you please add some comments, maybe some examples?

paulfouquet (changes requested):
Just some comments needed for understanding.
Otherwise looks good

paulfouquet (dismissed):

paulfouquet (approved):

PR #573, opened 2023-07-30
fix: standardise local files TDE-821

amfage:


scripts/cli/cli_helper.py
```diff
@@ -41,13 +41,20 @@ def format_source(source: List[str]) -> List[TileFiles]:
...
-    for s in source:
-        local_tile_file.append(TileFiles(output=os.path.splitext(get_file_name_from_path(s))[0], input=[s]))
-
-    return local_tile_file
+            raise Exception(e) from e
+    else:
+        source_dev: List[TileFiles] = []
+        for s in source:
+            gdalinfo: GdalInfo = gdal_info(s, False)
+            if scale > 0:
```
paulfouquet, line 49: What happen to `origin` if `not scale > 0`?

scripts/cli/cli_helper.py
```diff
@@ -1,24 +1,23 @@
...
+from scripts.tile.tile_index import TileIndexException, get_tile_name
 
 
 class TileFiles(NamedTuple):
     output: str
     input: List[str]
 
 
-def format_source(source: List[str]) -> List[TileFiles]:
+def get_tile_files(source: List[str], scale: int = 0) -> List[TileFiles]:
```
paulfouquet, line 18: The input parameter `source` is confusing in this function as it could be `["[{'output': 'CE16_5000_1001', 'input': ['s3://bucket/SN9457_CE16_10k_0501.tif']}]"]` or `['SN9457_CE16_10k_0501.tif', 'SN9457_CE16_10k_0502.tif']`.
I'm wondering if we should prepare the data separately so this function have only one kind of input?

scripts/standardise_validate.py
```diff
@@ -36,21 +31,27 @@ def main() -> None:
...
     )
+    parser.add_argument("--scale", dest="scale", help="Tile grid scale to align output tile to", required=True)
     parser.add_argument("--target", dest="target", help="Target output", required=True)
-    arguments = parser.parse_args()
 
-    source = arguments.source
-    from_file = arguments.from_file
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("--from-file", dest="from_file", help="The path to a json file containing the input tiffs")
+    group.add_argument("--dev", dest="dev", nargs="+", help="Developer mode - path to local files separated by spaces")
```
paulfouquet, line 39: I think positional arguments would be more appropriate than a `--dev` flag.

scripts/cli/cli_helper.py
paulfouquet: Also the input source of type `["[{'output': 'CE16_5000_1001', 'input': ['s3://bucket/SN9457_CE16_10k_0501.tif']}]"]` should probably not exist anymore as we would use `--from-file` with Argo.

scripts/tile/tile_index.py
```diff
@@ -33,6 +102,98 @@ class Bounds(NamedTuple):
     size: Size
 
 
+def round_with_correction(value: Union[int, float]) -> int | float:
```
paulfouquet, line 105: Should bring back the unit tests for that too

scripts/cli/cli_helper.py
amfage: Yeah, nothing - I'll refactor this.

amfage:
Will look again at local file input when the work in [TDE-820](https://toitutewhenua.atlassian.net/browse/TDE-820) and [TDE-828](https://toitutewhenua.atlassian.net/browse/TDE-828) are done.

PR #580, opened 2023-08-02
feat: download files using multithreading TDE-822

MDavidson17:


scripts/files/fs.py
```diff
@@ -42,3 +49,55 @@ def exists(path: str) -> bool:
...
+            write(f"{target.split('.')[0]}{ext}", read(f"{file.split('.')[0]}{ext}"))
+            get_log().info(
+                "Download tiff sidecars", path=f"{file.split('.')[0]}{ext}", target_path=f"{target.split('.')[0]}{ext}"
+            )
+        except:  # pylint: disable-msg=bare-except
+            pass
+    return download_path
+
+
+def download_tiffs_parallel_multithreaded(inputs: List[str], target: str, concurrency: int = 10) -> List[str]:
```
paulfouquet, line 79: `fs.py`  has been designed to provide generic functions to deal files within multiple file systems (currently local and s3). It could be good idea to allow multithreading while doing operations on multiple files. However, I would make it really generic (something like a `write_all()`) writting a list of files in a target as you do with `download_tiffs_parallel_multithreaded()`.
I think that the specificity of downloading tiffs and sidecar files should not be handled in here.

MDavidson17: I agree and I did struggle to figure out where to put this function, if you have a suggestion I would really appreciate it!
I would like to pull it out of standarding.py to make it more visisble because I think it could be resused and others might not find it if it is burried an unrelated file. 
Maybe it needs its own file in `scripts/files/`?

This is why I put it in `fs.py` but if we only care about the first option I am happy to move it to `fs_s3.py`
In `fs.py`:

1. read s3 - write local
2. read s3 - write s3
3. read local - write local (uncessary?)
4. read local - write s3 (probably not the best performance)

in `fs_s3.py`:

1. read s3 - write local (have to import fs_local.write which (to me) seems not right)
2. ~~read s3 - write s3 (to get this and above, have to import fs.write - circular import error)~~
3. ~~read local - write local (not possible but also uncessary?)~~
4. ~~read local - write s3 (not possible probably not the best performance)~~

paulfouquet: Having it in `fs.py` rather than `fs_s3.py` makes more sense as `fs.write` and `fs.read` already manage multiple file systems (local and s3) so you don't have to worry about them

scripts/files/fs.py
```diff
@@ -42,3 +48,64 @@ def exists(path: str) -> bool:
     if is_s3(path):
         return fs_s3.exists(path)
     return fs_local.exists(path)
+
+
+def _read_write_file(target: str, file: str) -> str:
```
paulfouquet, line 53: You could get rid of this function doing something like the suggestion below

scripts/files/fs.py
```diff
@@ -42,3 +48,64 @@ def exists(path: str) -> bool:
...
+    Args:
+        inputs: list of files to read
+        target: target folder to write to
+
+    Returns:
+        list of written file paths
+    """
+    written_tiffs: List[str] = []
+    with ThreadPoolExecutor(max_workers=concurrency) as executor:
+        futuress = {executor.submit(_read_write_file, target, input): input for input in inputs}
```
paulfouquet, line 81:
```suggestion
        futuress = {executor.submit(write, os.path.join(target, f"{input.split('/')[-1]}"), read(input)): input for input in inputs}
```

scripts/files/fs.py
```diff
@@ -42,3 +48,64 @@ def exists(path: str) -> bool:
...
+        for future in as_completed(futuress):
+            if future.exception():
+                get_log().warn("Failed Read-Write", error=future.exception())
+            else:
+                written_tiffs.append(future.result())
+
+    if len(inputs) != len(written_tiffs):
+        get_log().error("Missing Files", missing_file_count=len(inputs) - len(written_tiffs))
```
paulfouquet, line 89: You can simplify `missing_file_count` to `count` as we know the context with the error message

paulfouquet (commented):
A couple of suggestions.

scripts/files/fs.py
paulfouquet: (you might need to return the target path in `write()`)

scripts/files/fs.py
```diff
@@ -42,3 +49,50 @@ def exists(path: str) -> bool:
...
+        inputs: list of input files to search for extensions
+        extensions: the sidecar file extensions
+
+    Returns:
+        list of existing sidecar files
+    """
+    sidecars = []
+    for file in inputs:
+        for extension in extensions:
+            sidecar = f"{file.split('.')[0]}{extension}"
```
paulfouquet, line 95: You'll get into an issue if `file` value contains multiple `.`. Using os.path is safer when dealing with paths:
```python
>>> import os
>>> file = "my.file.txt"
>>> f"{file.split('.')[0]}.doc"
'my.doc'
>>> file_prefix, ext = os.path.splitext(file)
>>> f"{file_prefix}.doc"
'my.file.doc'
```

paulfouquet (changes requested):

scripts/files/fs.py
MDavidson17: Awesome, thanks was looking into this too, a case of only testing with s3 paths beforehand :facepalm: 
I noticed exists was also quite slow and detracted from the speed increase of downloading the files so I made this function multithreading too.

paulfouquet: Do we have some runtimes to compare?

- with/without multithreading download
- same with sidecar files

scripts/files/fs.py
```diff
@@ -42,3 +49,62 @@ def exists(path: str) -> bool:
...
+        inputs: list of files to read
+        target: target folder to write to
+
+    Returns:
+        list of written file paths
+    """
+    written_tiffs: List[str] = []
+    with ThreadPoolExecutor(max_workers=concurrency) as executor:
+        futuress = {
+            executor.submit(write, os.path.join(target, f"{os.path.basename(input)}"), read(input)): input for input in inputs
```
blacha, line 67: does this stream the file down or does it try and read all the files into memory and then write them out?

scripts/files/fs.py
MDavidson17: Here are some results for downloading the sidecars (average of 10 runs):
|                  | Before Changes                  | Multithread 1           | Multithread 25           |
|------------------|---------------------------------|-------------------------|--------------------------|
| Sidecar Download | 48s (download without checking) | 1min 23s (using exists) | 33s (multithread exists) |

I found it hard to test write_all but I re-ran waikato DEM and the [workflow using multithreading](https://argo.linzaccess.com/workflows/argo/test-mdavidson-multi-hb88v?tab=summary) took a total of 15mins rather than the [original without multithreading](https://argo.linzaccess.com/workflows/argo/waikato-2021-dem-1m2l7vx?tab=workflow) wich took 31 minutes.

My concern is that it still has the same `could not connect to endpoint url` error maybe somewhat more so with multithreading?

scripts/files/fs.py
MDavidson17: After our discussion I have reduced the concurrecy to 4 for now, I also added a comment to [TDE-827](https://toitutewhenua.atlassian.net/browse/TDE-827) to work on this as a part of the review/refactor of our repos.

amfage (approved):

paulfouquet (approved):

PR #595, opened 2023-08-09
feat: script to convert ascii files TDE-814

MDavidson17:
(Built on top of multithreading branch will rebase onto master and mark ready for review when that [PR](https://github.com/linz/topo-imagery/pull/580) is merged)

Includes some refactoring of existing listing/downloading code to avoid duplication:
- `list_json_in_uri` -> `list_uri` 

Has successfully run on a edited workflow in argo:
image


scripts/translate_ascii.py
```diff
@@ -0,0 +1,59 @@
...
+            asc_files = write_all(inputs=asc_files, target=tmp_path, concurrency=25)
+
+        else:
+            # list ascii files from local
+            asc_files = [os.path.join(source, file) for file in os.listdir(os.path.abspath(source)) if file.endswith(".asc")]
+
+        tiffs = []
+        for asc in asc_files:
+            # translate from ascii to geotiff using GDAL
+            tiff = os.path.join(tmp_path, f"{asc.split('/')[-1].split('.')[0]}.tiff")
```
paulfouquet, line 47: I would recommend using `os.path` to change the file extension name.

scripts/translate_ascii.py
```diff
@@ -0,0 +1,59 @@
...
+            command = ["gdal_translate", "-of", "GTiff", asc, tiff]
```
blacha, line 48: What does this output as? do we need a `-co COMPRESS=lzw` `-co NUM_THREADS=all_cpus` etc?

scripts/translate_ascii.py
```diff
@@ -0,0 +1,59 @@
...
+    --source - local or s3 path to source data
+    --target - local or s3 path to write converted tiffs
+
+    examples:
+        python translate_ascii.py --source s3://linz-elevation-staging/test/sample-tests/ --target /tmp/
+        python translate_ascii.py --source ./tests/data/ --target /tmp/
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--source", dest="source", required=True, help="S3 path to the input ascii files")
```
paulfouquet, line 24: I'm not sure to have a full understanding of this integrated inside a workflow.
Why not passing a path to a file (json) that contains the actual list of the ascii files. So you can exclude the logic of filtering the source directory (as we do in other scripts in this repo).

scripts/files/fs_s3.py
```diff
@@ -164,7 +163,7 @@ def prefix_from_path(path: str) -> str:
     return path.replace(f"s3://{bucket_name}/", "")
 
 
-def list_json_in_uri(uri: str, s3_client: Optional[boto3.client]) -> List[str]:
+def list_uri(uri: str, s3_client: Optional[boto3.client], extension: Optional[str] = None) -> List[str]:
     """Get the `JSON` files from a s3 path
```
paulfouquet, line 167: This description comment needs to be updated.

scripts/translate_ascii.py
paulfouquet, line 48: I would move the command declaration where we put the other ones.

scripts/files/fs_s3.py
MDavidson17: Base on your previous comment this change was no longer required so I reverted this function back to how it was

scripts/translate_ascii.py
MDavidson17: Good point, I have changed this to from-file and undated the temporary ascii workflow to use aws-list similar to how we structured for creating thumbnails

scripts/translate_ascii.py
MDavidson17: I've added both `-co COMPRESS=lzw` and `-co NUM_THREADS=all_cpus` this script runs before standarding so I was intending just to convert the ascii file to tiff and let standardising handle the rest?

blacha: Ideally we should never store anything uncompressed, converting from ASC to uncompressed tiff is a improvement, but still should atleast be `lzw` 

I thinkg there are a few standard parameters we need to always add to `gdal_translate` like the `NUM_THREADS` and defaulting to lzw for compression seems easy.

amfage (dismissed):

scripts/translate_ascii.py
```diff
@@ -0,0 +1,49 @@
...
+    asc_files = json.loads(read(arguments.from_file))
+
+    with tempfile.TemporaryDirectory() as tmp_path:
+        tiffs = []
+        for ls in asc_files:
+            for asc in ls:
+                # translate from ascii to geotiff using GDAL
+                filename = os.path.splitext(os.path.basename(asc))[0]
+                tiff = os.path.join(tmp_path, f"{filename}.tiff")
+                run_gdal(get_ascii_translate_command(), input_file=asc, output_file=tiff)
```
paulfouquet, line 37: It would be a good idea to run them in concurrency?

scripts/translate_ascii.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # copy any sidecar files to target
+        sidecars = []
+        for ls in asc_files:
+            sidecars += find_sidecars(ls, [".prj", ".tfw"])
+        write_all(inputs=sidecars, target=arguments.target)
+
+
+def translate_ascii(file: str, tmp_path: str) -> str:
+    # translate from ascii to geotiff using GDAL
```
paulfouquet, line 59: Would be better to be a docstring.

scripts/translate_ascii.py
```diff
@@ -0,0 +1,67 @@
...
+    filename = os.path.splitext(os.path.basename(file))[0]
```
paulfouquet, line 60: I would suggest using `get_file_name_from_path()`

scripts/translate_ascii.py
```diff
@@ -0,0 +1,67 @@
...
+from functools import partial
+from multiprocessing import Pool
+
+from scripts.cli.cli_helper import is_argo
+from scripts.files.fs import find_sidecars, read, write_all
+from scripts.gdal.gdal_helper import run_gdal
+from scripts.gdal.gdal_preset import get_ascii_translate_command
+
+
+def main() -> None:
```
paulfouquet, line 14: Would be nice to have some logs with stats like duration, number of files, etc.

paulfouquet (changes requested):

paulfouquet (approved):

PR #644, opened 2023-09-07
fix: re-standardising TileByteCounts error TDE-850

amfage:
For a 1:1 (non-retile) re-standardise of already processed imagery, the input filepath was the same as the output filepath.

### DIFF
--- a/.github/workflows/format-tests.yml
+++ b/.github/workflows/format-tests.yml
@@ -55,5 +55,5 @@ jobs:
 
       - name: End to end test - Restandardise Aerial Imagery
         run: |
-          docker run  -v ${HOME}/tmp/:/tmp/ topo-imagery python3 standardise_validate.py --from-file ./tests/data/restandardise.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01
-          cmp --silent ${HOME}/tmp/BG35_1000_4829.tiff ./scripts/tests/data/output/BG35_1000_4829.tiff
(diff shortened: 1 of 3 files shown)


paulfouquet:
> For a 1:1 (non-retile) re-standardise of already processed imagery, the input filepath was the same as the output filepath.
> 
> Will add a test for this in another PR.

Could add a test in this PR, I think standardise one of the `scripts/test/data/output/` files and make sure no error and the file does not change should verify the fix.

MDavidson17 (approved):

.github/workflows/format-tests.yml
```diff
@@ -52,3 +52,8 @@ jobs:
           docker run  -v ${HOME}/tmp/:/tmp/ topo-imagery python3 thumbnails.py --from-file ./tests/data/thumbnails.json --target /tmp/
           cmp --silent ${HOME}/tmp/CB07_GeoTifv1-02-thumbnail.jpg ./scripts/tests/data/output/CB07_GeoTifv1-02-thumbnail.jpg
           cmp --silent ${HOME}/tmp/CB07_TIFFv1-02-thumbnail.jpg ./scripts/tests/data/output/CB07_TIFFv1-02-thumbnail.jpg
+
+      - name: End to end test - Restandardise Aerial Imagery
+        run: |
+          docker run  -v ${HOME}/tmp/:/tmp/ topo-imagery python3 standardise_validate.py --from-file ./tests/data/restandardise.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01
```
paulfouquet, line 58: It's not standardising the file because it already exists (from standardising the Aerial one) . You need to either delete the old one or put it in another directory

paulfouquet (changes requested):

paulfouquet (approved):

PR #720, opened 2023-11-09
feat: lint GitHub Actions workflows TDE-919

l0b0:


pyproject.toml
```diff
@@ -54,4 +54,5 @@ pytest = "^7.4.3"
 pytest-dependency = "^0.5.1"
 moto = "^4.2.6"
 pytest-mock = "^3.12.0"
+shellcheck-py = "*"
```
paulfouquet, line 57: Where is that used? Is that something you run manually?

l0b0: It's an optional dependency of actionlint.

paulfouquet: How does that work? `actionlint` detects automatically that this dependency exists and run it?

l0b0: Exactly.

paulfouquet (approved):

PR #757, opened 2023-12-01
feat: title and description by arguments TDE-960

MDavidson17:
#### Motivation
Entering/typing a title or description for imagery datasets is currently inefficient and risky. To remove the risk the title & description should be automatically generated using input parameters.

#### Modification
collection_from_items.py updated to take: `region`, `gsd`, `subtype`, `location`, `start-datetime`, `end-datetime`, `event`, `historic-survey-number` and `lifecycle` parameters.

automation logic written in new file: `generate_metadata.py`

_NB: the tests provide a good over of the different title / description logic_ 

#### Checklist

- [x] Tests updated
- [x] Docs updated  - there are no docs explaining functions and parameters in topo-imagery - this does exist in argo-tasks so maybe we should add a ticket to do this?
- [x] Issue linked in Title

[PR Topo-workflows to be merged after this PR has been released](https://github.com/linz/topo-workflows/pull/323)
PR ADDING METADATA TO COLLECTION.JSON TO COME


scripts/collection_from_items.py
```diff
@@ -6,19 +6,35 @@
...
+    parser.add_argument("--location", dest="location", help="Optional Location of dataset, e.g.- Hutt City", required=False)
+    parser.add_argument(
+        "--start-datetime", dest="start_datetime", help="Start datetime in format YYYY-MM-DD", type=valid_date, required=True
+    )
+    parser.add_argument(
+        "--end-datetime", dest="end_datetime", help="End datetime in format YYYY-MM-DD", type=valid_date, required=True
+    )
+    parser.add_argument("--event", dest="dest", help="Event name if applicable", required=False)
+    parser.add_argument(
+        "--historic-survey-number", dest="historic_survey_number", help="Historic Survey Number if Applicable", required=False
```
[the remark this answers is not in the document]
MDavidson17: I have added an example but the naming is a question for @dwsilk maybe we could do `historic-survey-id`?

scripts/stac/imagery/generate_metadata.py
```diff
@@ -0,0 +1,176 @@
...
+    return None
+
+
+def _format_event_for_elevation_title(event: Optional[str]) -> Optional[str]:
+    if event and event != "":
+        return f"- {event}"
+    return None
+
+
+def _region_map(region: str) -> str:
```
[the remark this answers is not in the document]
MDavidson17: Sorry I not sure how this fits into this function, other than to edit the last line of the function but then its making this line more complicated?

Also maybe the naming of it should be changed to `_convert_region_ascii_to_utf8` ?

scripts/collection_from_items.py
```diff
@@ -6,19 +6,92 @@
...
+            "pacific-islands",
+            "southland",
+            "taranaki",
+            "tasman",
+            "waikato",
+            "wellington",
+            "west-coast",
+        ],
+    )
+    parser.add_argument("--gsd", dest="gsd", help="GSD of imagery Dataset", type=nullable_str, required=True)
```
l0b0, line 66: Is it really valid to specify `--gsd=''`, so that `args.gsd` ends up being `None`? If so, why not just make this optional?

scripts/stac/imagery/generate_metadata.py
```diff
@@ -0,0 +1,167 @@
...
+from typing import Optional
+
+
+class SubtypeParameterError(Exception):
+    def __init__(self, subtype: str) -> None:
+        self.message = f"Unrecognised/Unimplemented Subtype Parameter: {subtype}"
+
+
+class ImagerySubtypes(str, Enum):
+    SATELLIE = "Satellite Imagery"
```
l0b0, line 13: "SATELLIE" should be "SATELLITE".

scripts/collection_from_items.py
```diff
@@ -6,19 +6,92 @@
...
     parser.add_argument("--collection-id", dest="collection_id", help="Collection ID", required=True)
-    parser.add_argument("--title", dest="title", help="Collection title", required=True)
-    parser.add_argument("--description", dest="description", help="Collection description", required=True)
+    parser.add_argument(
+        "--subtype",
+        dest="subtype",
+        help="Dataset subtype description",
+        required=True,
+        choices=[
+            ImagerySubtypes.AERIAL,
```
l0b0, line 29: You can use `list(SomeEnum)` to get all the entries in an enum.

scripts/stac/imagery/generate_metadata.py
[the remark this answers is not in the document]
l0b0: Now that `--region` is an enum choice, you can get rid of all the logic and replacements by using a `dict` lookup:

```python
HUMAN_READABLE_REGIONS = {
    "hawkes-bay": "Hawke's Bay",
    …
}
```

Then, instead of having a `_region_map` you can just do `HUMAN_READABLE_REGIONS[region]`.

scripts/collection_from_items.py
MDavidson17: gsd is required for the title so, I have changed (locally atm) the workflow parameter to:
```
- name: gsd
description: 'Dataset GSD (e.g., 0.3m)'
```
If the user tries to submit an imagery-standardise workflow without specifying a gsd parameter it will fail. 
Therefore I have made args.gsd here a str

scripts/collection_from_items.py
```diff
@@ -6,19 +6,92 @@
...
+            "taranaki",
+            "tasman",
+            "waikato",
+            "wellington",
+            "west-coast",
+        ],
+    )
+    parser.add_argument("--gsd", dest="gsd", help="GSD of imagery Dataset", type=nullable_str, required=True)
+    parser.add_argument(
+        "--location", dest="location", help="Optional Location of dataset, e.g.- Hutt City", type=nullable_str, required=False
```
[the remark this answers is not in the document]
MDavidson17: So in argo, if I put in the workflow
```
- name: location
 description: '(Optional) Additional datatset description, to be used in the title in place of the Region, e.g.- Hamilton 0.05m Urban Aerial Photos (2023)'
```
The user has to enter a value otherwise they will get an error like:
` {"code":3,"message":"spec.arguments.location.value is required"}: spec.arguments.location.value is required`

In other places to get around this (i.e., to make an input parameter optional) we have used:
```
- name: location
 description: '(Optional) Additional datatset description, to be used in the title in place of the Region, e.g.- Hamilton 0.05m Urban Aerial Photos (2023)'
value: ''
```

I was under the impression that this will pass an empty string to create_collection.py instead of None.
But in this scenario `event=''` is the same as `event=None` right? so I have removed the nullable_str.

scripts/collection_from_items.py
MDavidson17: This didn't seem to work, I got the following errror:
```
collection_from_items.py: error: argument --subtype: invalid choice: 'Urban Aerial Photos' (choose from [<ImagerySubtypes.SATELLITE: 'Satellite Imagery'>, <ImagerySubtypes.UR
BAN: 'Urban Aerial Photos'>, <ImagerySubtypes.RURAL: 'Rural Aerial Photos'>, <ImagerySubtypes.AERIAL: 'Aerial Photos'>, <ImagerySubtypes.HISTORICAL: 'Scanned Aerial Photos'>]
, [<ElevationSubtypes.DEM: 'DEM'>, <ElevationSubtypes.DSM: 'DSM'>])
```

I changed it to: `choices=[type.value for type in ImagerySubtypes] + [type.value for type in ElevationSubtypes],`, this works, I'm uncertain if it is the best method for getting these values.

scripts/collection_from_items.py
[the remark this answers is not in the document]
l0b0: Oh right, you're working around Argo's lack of a way to not pass empty parameter values. Sounds good!

scripts/collection_from_items.py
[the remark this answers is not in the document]
MDavidson17: Asked Daniel, makes sense to keep it as number despite it being a string input as it is called number everywhere else i.e., SN stands for `Survey Number`.

Unfortuantely we can't take a number and append `SN` as sometimes it might be `SNC` or have an `A` or `B` at the end.

l0b0 (dismissed):

scripts/stac/imagery/generate_metadata.py
```diff
@@ -0,0 +1,176 @@
...
+    # pylint: disable-msg=too-many-arguments
+
+    _validate_no_empty_strings({"region": region, "subtype": subtype, "gsd": gsd})
+
+    date = _format_date_for_metadata(start_date, end_date)
+    name = _format_name_for_title(_region_map(region), location)
+    preview = _is_preview(lifecycle)
+
+    if historic_survey_number:
+        return " ".join(f"{name} {gsd} {historic_survey_number} ({date}) {preview or ''}".split())
```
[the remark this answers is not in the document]
blacha: This is a very interesting way of approaching this, had no idea that pythons split would automatically trim the split too, so you will lose \n \r as well! 

Might be a idea to add a comment for future reference?

scripts/stac/imagery/generate_metadata.py
```diff
@@ -0,0 +1,181 @@
...
+        start_date: Dataset capture start date
+        end_date: Dataset capture end date
+        lifecycle: Dataset status
+        location: Optional location of dataset, e.g.- Hutt City
+        event: Optional details of capture event, e.g. - Cyclone Gabreille
+        historic_survey_number: Optional historic imagery survey number, e.g.- SNC88445
+    Returns:
+        Dataset Title
+    """
+    # pylint: disable-msg=too-many-arguments
```
blacha, line 80: What is the pythonic way of handling too many arguments especially with similar "str" arguments it can be very easy to put them into the wrong order eg swap region for gsd?

l0b0: Adding `*` as a pseudo-parameter before all the others force users to pass arguments by name, making it much less likely to accidentally swap them. This is already considered best practice by some projects, since it eliminates a whole class of problems with very little overhead. From the [docs](https://docs.python.org/3.5/reference/compound_stmts.html#function-definitions):

> Parameters after “`*`” or “*`identifier`” are keyword-only parameters and may only be passed used keyword arguments.

MDavidson17: I have turned the optional args into **kwargs, this was enough to remove the pylint warning.
I left the required args as is because we want the function to fail if they are not provided?

l0b0: I'd rather revert and disable too-many-arguments globally than use this pattern - this is hiding information inside `kwargs`.

l0b0: That said, doesn't the `generate_title` and `generate_description` logic belong inside `ImageryCollection`?

MDavidson17: Sorry Victor somehow I missed your first comment I have added the pseudo * parameter.

I've left the kwargs but can change it back, seems like we have to pick and I'm not sure what the best/right option is..

I split this code into its own file as the collection file was already quite long but I can merge them together if that is preferred?

l0b0: `**kwargs` is appropriate if you don't know which arguments the function is going to be called with - you can't know what the parameter names should be. But this is not the case here - you know which keyword arguments are possible, and only ever handle those specifically. Which means the function signature is less strict than it could be.

l0b0: As discussed, *if* `generate_title` and `generate_description` are always run in the context of trying to create an `ImageryCollection`, it probably makes sense to instead pass their parameters to the `ImageryCollection` constructor, then *either* have the constructor set `self.title` and `self.description` itself, or have `@property`-annotated methods for both of them. Creating the title and description directly in the constructor (possibly via private methods) makes sense if the constructor receives all the information necessary to create them, that is, there's nothing which should ever modify the title or description outside of the constructor. Using `@property` methods make more sense if anything outside the constructor could cause the title or description to change.

MDavidson17: So I've:
- put all the user inputs into a typed dict which gets rid of the pylint warning and removes the kwargs
- Moved title and description into the ImageryCollection class as private methods, at this stage I don't expect these to change after being created so opted for that over `@property`

l0b0 (approved):

scripts/stac/imagery/metadata_constants.py
```diff
@@ -0,0 +1,69 @@
...
+class CollectionTitleMetadata(TypedDict):
+    """
+    region: Region of Dataset
+    gsd: Dataset Ground Sample Distance
+    start_date: Dataset capture start date
+    end_date: Dataset capture end date
+    lifecycle: Dataset status
+    Optional:
+        location: Optional location of dataset, e.g.- Hutt City
+        event: Optional details of capture event, e.g. - Cyclone Gabreille
```
amfage, line 15:
```suggestion
        event: Optional details of capture event, e.g. - Cyclone Gabrielle
```

scripts/stac/imagery/metadata_constants.py
```diff
@@ -0,0 +1,69 @@
...
+        historic_survey_number: Optional historic imagery survey number, e.g.- SNC88445
```
amfage, line 16:
```suggestion
        location: Optional location of dataset, e.g. Hutt City
        event: Optional details of capture event, e.g. Cyclone Gabrielle
        historic_survey_number: Optional historic imagery survey number, e.g. SNC88445
```

l0b0 (dismissed):

amfage (dismissed):

amfage (approved):

l0b0 (approved):

PR #758, opened 2023-12-01
feat: add capture-area.geojson to the STAC Collection TDE-965

paulfouquet:
#### Motivation

Based on https://github.com/linz/imagery/pull/273
Giving a capture area of the entire dataset to the consumer.

#### Modification

Add a `capture-area` in the STAC Collection.
Example 
[capture-area_wellington_2021_0.3m.zip](https://github.com/linz/topo-imagery/files/14118911/capture-area_wellington_2021_0.3m.zip)

A PR in https://github.com/linz/topo-workflows/ will be made to publish the `capture-area.geojson` along with the other files.

#### Checklist

- [x] Tests updated
- [ ] Docs updated N/A
- [x] Issue linked in Title


scripts/stac/imagery/tests/collection_test.py
```diff
@@ -156,3 +156,14 @@ def test_default_provider_missing() -> None:
...
+
+def test_capture_data_asset_present() -> None:
+    title = "Test Urban Imagery"
+    description = "Test Urban Imagery Description"
+    collection = ImageryCollection(title, description)
+
+    assert collection.stac["assets"]["capture_area"]["href"] == "./capture-area.geojson"
+    assert collection.stac["assets"]["capture_area"]["title"] == "Capture area"
+    assert collection.stac["assets"]["capture_area"]["type"] == "application/geo+json"
+    assert collection.stac["assets"]["capture_area"]["roles"] == ["metadata"]
```
blacha, line 169: where do we add the checksums and filesize?

How big is this geojson do we need to compress it?

paulfouquet: Apparently we can use the `file` fields: https://github.com/stac-extensions/file#asset--link-object-fields

I've added an example from `CastleRockQuarry` dataset in the PR description. It's about 400bytes.

scripts/standardising.py
```diff
@@ -189,5 +192,18 @@ def standardising(
         # Need GDAL to write to temporary location so no broken files end up in the done folder.
         run_gdal(command, input_file=input_file, output_file=standardized_working_path)
 
+        # Create footprint GeoJSON
+        run_gdal(
+            ["gdal_footprint", "-t_cs", "georef", "-t_srs", "+proj=longlat +ellps=WGS84"],
```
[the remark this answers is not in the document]
paulfouquet: `t_cs` [target coordinate system](https://gdal.org/programs/gdal_footprint.html#cmdoption-gdal_footprint-t_cs) I thought I would have to use that if we were using `WSG84`

> why is -t_srs a proj string rather than a epsg code?

Is that not how we should specify `WSG84`?

scripts/collection_from_items.py
```diff
@@ -19,7 +25,7 @@
 from scripts.stac.imagery.provider import Provider, ProviderRole
 
 
-# pylint: disable-msg=too-many-locals
+# pylint: disable-msg=too-many-locals,too-many-statements
```
l0b0, line 28: Looks like it might be time to split this up.

scripts/files/fs_s3.py
```diff
@@ -168,15 +167,16 @@ def prefix_from_path(path: str) -> str:
     return path.replace(f"s3://{bucket_name}/", "")
 
 
-def list_json_in_uri(uri: str, s3_client: Optional[client]) -> List[str]:
-    """Get the `JSON` files from a s3 path
+def list_files_in_uri(uri: str, suffixes: Tuple[str, ...], s3_client: Optional[client]) -> List[str]:
```
l0b0, line 170: Since `suffixes` is arbitrary length, while n-tuples are fixed length, it might be more appropriate to use either `List[str]` or even `Iterable[str]`.

scripts/collection_from_items.py
```diff
@@ -112,28 +118,51 @@ def main() -> None:
...
-                "skipping: item.collection != collection.id",
-                file=key,
-                action="collection_from_items",
-                reason="skip",
-            )
-            continue
-
-        collection.add_item(item_stac)
-        get_log().info("item added to collection", item=item_stac["id"], file=key)
+        content = json.loads(result["Body"].read().decode("utf-8"))
```
l0b0, line 128: I believe `json.load(result["Body"])` should work here, and should be more space-efficient since I expect it converts as it loads.

scripts/collection_from_items.py
```diff
@@ -112,28 +118,51 @@ def main() -> None:
...
+                    action="collection_from_items",
+                    reason="skip",
+                )
+                continue
+            collection.add_item(content)
+            get_log().info("item added to collection", item=content["id"], file=key)
+        elif key.endswith("_footprint.geojson"):
+            get_log().debug(f"adding geometry from {key}")
+            geom = shapely.geometry.shape(content["features"][0]["geometry"])
+            polygons.append(geom)
```
l0b0, line 144: I can't quite tell whether there should be an `else` here. Shouldn't we throw in that case?

scripts/collection_from_items.py
```diff
@@ -112,28 +118,51 @@ def main() -> None:
...
+
+    capture_area = geojson.Feature(geometry=shapely.ops.unary_union(polygons), properties={})
+    with TemporaryDirectory() as tmp_path:
+        tmp_capture_area_path = os.path.join(tmp_path, CAPTURE_AREA_FILE_NAME)
+        write(
+            tmp_capture_area_path,
+            json.dumps(capture_area).encode("utf-8"),
```
l0b0, line 151: Writing from a file descriptor should be more efficient than first reading it into memory. But I'm not sure if this sort of optimisation is worth it.

scripts/files/fs_s3.py
```diff
@@ -168,15 +167,16 @@ def prefix_from_path(path: str) -> str:
...
+    """Get a list of file paths from a s3 path based on their suffixes
 
     Args:
         uri: an s3 path
+        suffixes: a tuple of suffixes. example: (".json", "_meta.xml")
         s3_client: an s3 client
 
     Returns:
-        a list of JSON files
+        a list of file path
```
l0b0, line 179:
```suggestion
        a list of file paths
```

scripts/stac/imagery/collection.py
```diff
@@ -59,6 +63,32 @@ def __init__(
...
+
+        file_checksum = checksum.multihash_as_hex(path)
+        capture_area = {
+            "href": f"./{CAPTURE_AREA_FILE_NAME}",
+            "title": "Capture area",
+            "type": ContentType.GEOJSON,
+            "roles": ["metadata"],
+            "file:checksum": file_checksum,
+        }
+        self.stac["assets"]["capture_area"] = capture_area
```
l0b0, line 90: You can use [this pattern](https://stackoverflow.com/a/17756005/96588) to avoid the `if not self.stac.get("asset"):` above:

```python
>>> stac = {}
>>> stac.setdefault("assets", {})["capture_area"] = "foo"
>>> stac
{'assets': {'capture_area': 'foo'}}
```

scripts/collection_from_items.py
paulfouquet: We are listing only files that are `.json` or `_footprint.geojson`. If there was another kind of file that would mean `list_files_in_uri()` has a problem?

l0b0: OK, so this duplicates part of the functionality of `list_files_in_uri`. In that case, how much work would it be to refactor so that we don't have to check the suffix again in this function? For example, we could change `list_files_in_uri` to return a *tuple* of lists of strings corresponding to the files matching each of the `suffixes`, then handle those separately.

paulfouquet: As discussed, it won't be optimal to return different lists per suffix as we would need to call `get_object_parallel_multithreading` multiple time rather than once.

l0b0 (approved):

scripts/stac/imagery/collection.py
```diff
@@ -59,6 +68,47 @@ def __init__(
...
+    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str) -> None:
+        """Add the capture area of the Collection.
+        The `href` or path of the capture-area.geojson is always set as the relative `./capture-area.geojson`
+
+        Args:
+            polygons: list of geometries
+            target: path of the capture-area-geojson file
+        """
+
+        capture_area = geojson.Feature(geometry=shapely.ops.unary_union(polygons), properties={})
```
blacha, line 80: if polygons are slightly miss aligned eg 15.999997 and 16.00001 does this actually join the polygons?

scripts/stac/imagery/collection.py
```diff
@@ -59,6 +68,47 @@ def __init__(
...
+                content_type=ContentType.GEOJSON.value,
+            )
+
+            file_checksum = checksum.multihash_as_hex(tmp_capture_area_path)
+            capture_area = {
+                "href": f"./{CAPTURE_AREA_FILE_NAME}",
+                "title": "Capture area",
+                "type": ContentType.GEOJSON,
+                "roles": ["metadata"],
+                "file:checksum": file_checksum,
```
blacha, line 95: would be good to include "file:size" too

scripts/stac/imagery/collection.py
paulfouquet: I haven't test that. We should validate this with a unit test and maybe have a chat about it.

blacha: as we are projecting things from 2193->4326 the numbers may vary slightly

scripts/collection_from_items.py
```diff
@@ -115,28 +118,37 @@ def main() -> None:
...
-            continue
-
-        collection.add_item(item_stac)
-        get_log().info("item added to collection", item=item_stac["id"], file=key)
+        content = json.load(result["Body"])
+        # The following if/else looks like it could be avoid by refactoring `list_files_in_uri()`
+        # to return a result list per suffix, but we would have to call `get_object_parallel_multithreading()`
+        # for each of them to avoid this if/else.
+        if key.endswith(SUFFIX_JSON):
+            if not arguments.collection_id == content.get("collection"):
```
l0b0, line 133: `not` is usually used for things like `x not in y`; personally I find it clearer when there's a single comparator rather than a comparator and a negation:

```suggestion
            if arguments.collection_id != content.get("collection"):
```

scripts/files/tests/fs_s3_test.py
```diff
@@ -124,3 +124,20 @@ def test_exists_object_starting_with_not_exists() -> None:
     file_exists = exists("s3://testbucket/hello/another.fi")
 
     assert file_exists is False
+
+
+@mock_s3  # type: ignore
+def test_list_files_in_uri() -> None:
```
l0b0, line 130: A [slightly different pattern](https://github.com/linz/geostore/blob/master/tests/test_import_dataset.py#L84) might be useful to get type annotations for your S3 clients.

scripts/files/tests/fs_s3_test.py
```diff
@@ -124,3 +124,20 @@ def test_exists_object_starting_with_not_exists() -> None:
...
+    s3 = resource("s3", region_name=DEFAULT_REGION_NAME)
+    boto3_client = client("s3", region_name=DEFAULT_REGION_NAME)
+    s3.create_bucket(Bucket="testbucket")
+    boto3_client.put_object(Bucket="testbucket", Key="data/collection.json", Body=b"test content")
```
l0b0, line 134: The bucket name should probably be pulled out into a variable, to make sure it's clear that it's a re-used value.

As for the `Body`, we don't actually care what the contents are, so we could either use an `any_body()`, `ANY_BODY`, or an empty `b""`.

scripts/files/tests/fs_s3_test.py
```diff
@@ -124,3 +124,20 @@ def test_exists_object_starting_with_not_exists() -> None:
...
+    boto3_client.put_object(Bucket="testbucket", Key="data/image.tiff", Body=b"test content")
+    boto3_client.put_object(Bucket="testbucket", Key="data/image_meta.xml", Body=b"test content")
+
+    files = list_files_in_uri("s3://testbucket/data/", [".json", "_meta.xml"], boto3_client)
+
+    assert len(files) == 2
+    assert "data/collection.json" in files
+    assert "data/image_meta.xml" in files
+    assert "data/image.tiff" not in files
```
l0b0, line 143: Could we simply

```suggestion
    assert files == ["data/collection.json", "data/image_meta.xml"]
```

or, if the sequence is actually arbitrary,

```suggestion
    assert set(files) == {"data/collection.json", "data/image_meta.xml"}
```

scripts/stac/imagery/capture_aera.py
```diff
@@ -0,0 +1,68 @@
...
+from shapely import BufferCapStyle, BufferJoinStyle, Geometry, to_geojson, union_all
+from shapely.geometry import Polygon
+
+DECIMAL_DEGREES_1M = 0.00001
+"""
+Degree precision of ~1m (decimal places 5, https://en.wikipedia.org/wiki/Decimal_degrees)
+"""
+
+
+def to_feature(geometry: Geometry) -> Dict[str, Any]:
```
l0b0: Is there a GeoJSON `Feature` type we could use as the return type, or possibly a `TypedDict` to avoid the `Any`?

scripts/stac/imagery/capture_aera.py
```diff
@@ -0,0 +1,68 @@
...
+    """Transform a Geometry to a GeoJSON feature.
+
+    Args:
+        geometry: a Geometry
+
+    Returns:
+        a GeoJSON document.
+    """
+    return {"geometry": json.loads(to_geojson(geometry)), "type": "Feature", "properties": {}}
```
l0b0: Do we need to specify the `properties` property?

scripts/stac/imagery/capture_aera.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    buffered_polygons = []
+    for poly in polygons:
+        # Buffer each polygon to round up to the `buffer_distance`
+        buffered_poly = poly.buffer(buffer_distance, cap_style=BufferCapStyle.flat, join_style=BufferJoinStyle.mitre)
+        buffered_polygons.append(buffered_poly)
+    union = union_all(buffered_polygons)
+    # Negative buffer back in the polygons
+    union = union.buffer(-buffer_distance, cap_style=BufferCapStyle.flat, join_style=BufferJoinStyle.mitre)
+    union = union.simplify(buffer_distance)
```
l0b0: Could we come up with some different names for the three `union`s to clarify the intermediate results? Maybe something like `raw_union`, `unbuffered_union`, and `simplified_union`?

scripts/stac/imagery/collection.py
```diff
@@ -75,6 +84,50 @@ def __init__(
...
+                "file:size": file_stats.st_size,
+            }
+            self.stac.setdefault("assets", {})["capture_area"] = capture_area
+
+            # Save `capture-area.geojson` in target
+            write(
+                os.path.join(target, CAPTURE_AREA_FILE_NAME),
+                read(tmp_capture_area_path),
```
l0b0, line 121: Would it be possible to hold `json.dumps(capture_area_content).encode("utf-8")` in memory rather than writing it to a temporary file and then reading it back out again?

scripts/stac/imagery/collection.py
```diff
@@ -75,6 +84,50 @@ def __init__(
...
+                content_type=ContentType.GEOJSON.value,
+            )
+
+        if not self.stac.get("stac_extensions"):
+            self.stac["stac_extensions"] = []
```
l0b0, line 126: You can make the conditional implicit by doing this:

```suggestion
        self.stac["stac_extensions"] = self.stac.get("stac_extensions", [])
```

scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -0,0 +1,144 @@
...
+    # The following polygon is off by 0.1 to the "right" from the previous one
+    polygons.append(Polygon([(1.1, 1.0), (2.0, 1.0), (2.0, 0.0), (1.0, 0.0), (1.1, 1.0)]))
+    expected_merged_polygon = Polygon([(0.0, 1.0), (2.0, 1.0), (2.0, 0.0), (0.0, 0.0), (0.0, 1.0)])
+    print(f"GeoJSON expected: {to_feature(expected_merged_polygon)}")
+
+    # Here, the allowed margin of error is 0.01 < 0.1
+    merged_polygons = merge_polygons(polygons, 0.01)
+    print(f"GeoJSON result: {to_feature(merged_polygons)}")
+
+    assert not merged_polygons.equals(expected_merged_polygon)
```
l0b0, line 51: Would it be possible to assert what the value ends up being, rather than what it's not? Basically, there are an infinite number of polygons it's *not* equal to, so if there's any typos in `expected_merged_polygon` it would succeed for the wrong reason.

scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def test_generate_capture_area_rounded() -> None:
+    polygons = []
+    polygons.append(
+        Polygon(
+            [
+                (178.254184985672936, -38.40856551170436),
```
l0b0, line 59: Would it be possible to simplify these polygons? I can't tell whether the test is correct because of the complexity of the inputs.

scripts/stac/imagery/tests/collection_test.py
```diff
@@ -177,6 +179,65 @@ def test_default_provider_is_present(metadata: CollectionMetadata) -> None:
...
+    assert StacExtensions.file.value in collection.stac["stac_extensions"]
+    assert "file:checksum" in collection.stac["assets"]["capture_area"]
+    assert (
+        collection.stac["assets"]["capture_area"]["file:checksum"]
+        == "12209043a7059c39aa8a8bd22ad0e50476b1628a5b20df11c5a507900cab19f2741b"
+    )
+    assert "file:size" in collection.stac["assets"]["capture_area"]
+    assert collection.stac["assets"]["capture_area"]["file:size"] == 341
+    rmtree(target)
```
l0b0, line 238: We should probably use `with mkdtemp() as target` to remove the directory if the test fails. Or do you want to keep it?

scripts/standardising.py
```diff
@@ -200,7 +204,18 @@ def standardising(
 
         with TiffFile(standardized_working_path) as file_handle:
             if any(tile_byte_count != 0 for tile_byte_count in file_handle.pages.first.tags["TileByteCounts"].value):
+                # Create footprint GeoJSON
+                run_gdal(
+                    ["gdal_footprint", "-t_srs", "EPSG:4326"],
```
l0b0, line 209: It would help to have a constant with a name explaining what `"EPSG:4326"` is.

scripts/stac/imagery/capture_aera.py
paulfouquet: > A Feature object has a member with the name "properties".
https://datatracker.ietf.org/doc/html/rfc7946#section-3.2

The "has" seems a mandatory thing to me. But I could be wrong.

scripts/stac/imagery/capture_aera.py
paulfouquet: I spent some time looking at this today:
- there is a `geojson` library that I've been using in a previous commit in this PR but it does round the coordinates and I could not find a way to avoid it
- Creating a TypedDict seems very complex (managing all kind of `coordinates` field that are different for each Geometry Type for example).
I think at the moment it will make the code over-complicated, so I pass on this.

scripts/files/tests/fs_s3_test.py
paulfouquet: Would be better to do that in a separate PR.

scripts/files/tests/fs_s3_test.py
```diff
@@ -124,3 +124,20 @@ def test_exists_object_starting_with_not_exists() -> None:
...
+def test_list_files_in_uri() -> None:
+    bucket_name = "testbucket"
+    s3 = resource("s3", region_name=DEFAULT_REGION_NAME)
+    boto3_client = client("s3", region_name=DEFAULT_REGION_NAME)
+    s3.create_bucket(Bucket=bucket_name)
+    boto3_client.put_object(Bucket=bucket_name, Key="data/collection.json", Body=b"")
+    boto3_client.put_object(Bucket=bucket_name, Key="data/image.tiff", Body=b"")
+    boto3_client.put_object(Bucket=bucket_name, Key="data/image_meta.xml", Body=b"")
+
+    files = list_files_in_uri("s3://testbucket/data/", [".json", "_meta.xml"], boto3_client)
```
l0b0, line 139: We should probably use `bucket_name` here:

```suggestion
    files = list_files_in_uri(f"s3://{bucket_name}/data/", [".json", "_meta.xml"], boto3_client)
```

scripts/stac/imagery/capture_aera.py
l0b0: > The value of the properties member is an object (any JSON object or a JSON null value).

Seems pretty clear it's mandatory if it can have a `null` value. :+1:

scripts/stac/imagery/capture_area.py
```diff
@@ -0,0 +1,70 @@
...
+    Returns:
+        The capture-area geojson document.
+    """
+    # It allows to round the geometry as we've seen some tiffs geometry being slightly off,
+    # sometimes due to rounding issue in their creation process (before delivery).
+    # If we don't apply this rounding, we could get a very small gaps between tiffs
+    # which would result in a capture area having gaps.
+    # We multiply the gsd (in meter) to the 1m degree of precision.
+    # Note that all the polygons are buffered which means a gap bigger than the gsd,
+    # but < gsd*2) will be closed.
```
l0b0, line 66: Does this text possibly belong in the docstring? Multiple single-line comments like this are not docstrings, and are not treated as a single block comment.

scripts/stac/imagery/collection.py
```diff
@@ -75,6 +83,41 @@ def __init__(
...
+    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str) -> None:
+        """Add the capture area of the Collection.
+        The `href` or path of the capture-area.geojson is always set as the relative `./capture-area.geojson`
+
+        Args:
+            polygons: list of geometries
+            target: path of the capture-area-geojson file
+        """
+
+        # The GSD is currently alway in meters (e.g. `0.3m`)
```
l0b0, line 95: Why "currently"? Is it going to be a different unit in the future? If not, we should probably avoid that phrasing

Also, typo, punctuation after "e.g.". Suggestion:

```suggestion
        # The GSD is measured in meters (e.g., `0.3m`)
```

l0b0 (dismissed):

l0b0 (approved):

l0b0 (approved):

PR #771, opened 2023-12-11
refactor: PEP-8 compliance

l0b0:
#### Motivation

Consistency and PEP-8 compliance.

#### Modification

Found some issues with IDEA.

#### Checklist

None of these are applicable.

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/scripts/gdal/gdal_bands.py
+++ b/scripts/gdal/gdal_bands.py
@@ -28,6 +28,7 @@ def get_gdal_band_offset(file: str, info: Optional[GdalInfo] = None, preset: Opt
     Args:
         file: file to check
         info: optional precomputed gdalinfo
+        preset: "dem_lerc" preset used to differentiate single band elevation tiffs from single band historical imagery
 
     Returns:
         list of band mappings eg "-b 1 -b 1 -b 1"
--- a/scripts/aws/aws_helper.py
+++ b/scripts/aws/aws_helper.py
@@ -4,9 +4,9 @@ from time import sleep
 from typing import Any, Dict, List, NamedTuple, Optional
 from urllib.parse import urlparse
 
-import boto3
-import botocore
+from boto3 import Session
 from botocore.credentials import AssumeRoleCredentialFetcher, DeferredRefreshableCredentials, ReadOnlyCredentials
+from botocore.session import Session as BotocoreSession
 from linz_logger import get_log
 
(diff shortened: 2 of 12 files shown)


blacha:
:lgtm: did you want to change the title to something other than "misc fixes" how about something like

"refactor: pep-8 compliance" ?

scripts/gdal/gdal_bands.py
```diff
@@ -28,6 +28,7 @@ def get_gdal_band_offset(file: str, info: Optional[GdalInfo] = None, preset: Opt
     Args:
         file: file to check
         info: optional precomputed gdalinfo
+        preset: "dem_lerc" if single band (?!?)
```
paulfouquet, line 31: ?!?

l0b0: Yeah, I couldn't work out what the parameter was supposed to do. The only value which is handled in the code is `"dem_lerc"`, which I couldn't find anywhere online. So it might as well be a boolean flag.

MDavidson17: presets are picked from an enum in [topo-workflows](https://github.com/linz/topo-workflows/blob/1e3a71cdeacfdbef1ed8e9e42b01a8ee02aa0f05/workflows/raster/standardising.yaml#L93C10-L93C10) which somewhat confusingly is called compression - also I think we made up the specific term `dem_lerc`.

I think their intended use was for [constructing the gdal commands](https://github.com/linz/topo-imagery/blob/250b82c196c1f3db601e9712d6834d6da28e6952/scripts/gdal/gdal_preset.py#L36). However, when we were updating the code to work for DEM/DSM data we needed a way to differentiate between singleband elevation tiffs and single band historical imagery, as we already had the dem_lerc preset we used that. 

I was thinking now that we are taking a category and preset values in argo there might be an opportunity to tidy up how we differentiate between aerial imagery and elevation data throughout this repo - but that would be its own task and PR.

l0b0: Thank you! Do you have a suggestion for the comment for now?

MDavidson17: ahh you could workshop something along the lines of: `dem_lerc preset used to differentiate single band elevation tiffs from single band historical imagery`

MDavidson17 (approved):

amfage (approved):

PR #835, opened 2024-02-02
refactor: title & description TDE-955

MDavidson17:
#### Motivation

refactor title & description formatting to align with naming documentation

#### Modification

- docstring updated to point to `naming.md`
- `event` removed from all titles as should now also be included in `geographic_description`
- `geographic_description` removed from elevation description otherwise, if the dataset has an event it will be repeated

Examples:
`Hawke's Bay Cyclone Gabrielle 0.3m Rural Aerial Photos (2023)`
`Ponsonby 0.3m Rural Aerial Photos (2023)`

`Hawke's Bay Cyclone Gabrielle 0.3m Satellite Imagery (2023)`

`Auckland LiDAR 0.3m DSM (2023) - Preview`

#### Checklist

- [x] Tests updated
- [x] Docs updated - just docstrings
- [x] Issue linked in Title

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -210,12 +210,10 @@ class ImageryCollection:
 
     def _title(self) -> str:
         """Generates the title for imagery and elevation datasets.
-        Satellite Imagery / Urban Aerial Photos / Rural Aerial Photos:
+        Satellite Imagery / Urban Aerial Photos / Rural Aerial Photos / Scanned Aerial Photos:
           https://github.com/linz/imagery/blob/master/docs/naming.md
         DEM / DSM:
           https://github.com/linz/elevation/blob/master/docs/naming.md
-        If Historic Survey Number:
-          [geographic_description / Region] [GSD] [Survey Number] ([Year(s)]) [?- Preview]
         Returns:
             Dataset Title
         """
@@ -224,27 +222,27 @@ class ImageryCollection:
         historic_survey_number = self.metadata.get("historic_survey_number")
 
         # format date for metadata
+        date = f"{self.metadata['start_datetime'].year}-{self.metadata['end_datetime'].year}"
         if self.metadata["start_datetime"].year == self.metadata["end_datetime"].year:
             date = str(self.metadata["start_datetime"].year)
-        else:
-            date = f"{self.metadata['start_datetime'].year}-{self.metadata['end_datetime'].year}"
 
         # determine dataset name
+        region = HUMAN_READABLE_REGIONS[self.metadata["region"]]
+        imagery_name = region
+        elevation_description = None
         if geographic_description:
-            name = geographic_description
-        else:
-            name = HUMAN_READABLE_REGIONS[self.metadata["region"]]
+            imagery_name = geographic_description
+            elevation_description = f"- {geographic_description}"
 
         # determine if dataset is preview
+        preview = None
         if self.metadata.get("lifecycle") == "preview":
             preview = "- Preview"
-        else:
-            preview = None
 
         if self.metadata["category"] == SCANNED_AERIAL_PHOTOS:
             if not historic_survey_number:
                 raise MissingMetadataError("historic_survey_number")
-            return " ".join(f"{name} {self.metadata['gsd']} {historic_survey_number} ({date}) {preview or ''}".split())
+            return " ".join(f"{imagery_name} {self.metadata['gsd']} {historic_survey_number} ({date}) {preview or ''}".split())
 
         if self.metadata["category"] in [
             SATELLITE_IMAGERY,
@@ -252,11 +250,11 @@ class ImageryCollection:
             RURAL_AERIAL_PHOTOS,
         ]:
             return " ".join(
-                f"{name} {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
+                f"{imagery_name} {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
             )
         if self.metadata["category"] in [DEM, DSM]:
             return " ".join(
-                f"{name} LiDAR {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
+                f"{region} {elevation_description or ''} LiDAR {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
             )
         raise SubtypeParameterError(self.metadata["category"])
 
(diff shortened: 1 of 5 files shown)


scripts/stac/tests/generate_title_test.py
```diff
@@ -103,27 +103,30 @@ def test_generate_title_geographic_description(metadata: Tuple[CollectionMetadat
...
 
 
 def test_generate_title_event_elevation(metadata: Tuple[CollectionMetadata, CollectionMetadata]) -> None:
     _, metadata_hb = metadata
     metadata_hb["category"] = "dsm"
+    metadata_hb["geographic_description"] = "Hawke's Bay Cyclone Gabrielle"
     metadata_hb["event_name"] = "Cyclone Gabrielle"
     collection = ImageryCollection(metadata_hb)
-    title = "Hawke's Bay - Cyclone Gabrielle LiDAR 0.3m DSM (2023)"
+    title = "Hawke's Bay - Hawke's Bay Cyclone Gabrielle LiDAR 0.3m DSM (2023)"
```
MDavidson17, line 119: This doesn't look nice... I could change it to:
```python
if event:
    title = "region - event LiDAR..."
   # Hawke's Bay - Cyclone Gabrielle LiDAR 1m DSM (2023)
   # Nelson - Top of the South Floods LiDAR 1m DSM (2023)
elif geographic_description:
    title = "region - geographic_description LiDAR..."
    # Wellington - Hutt City LiDAR 1m DEM (2022)
title = "region LiDAR..."
# Taranaki LiDAR 1m DEM (2020)
```

scripts/stac/imagery/collection.py
```diff
@@ -210,94 +210,70 @@ def write_to(self, destination: str) -> None:
...
         """
         # format optional metadata
         geographic_description = self.metadata.get("geographic_description")
         historic_survey_number = self.metadata.get("historic_survey_number")
-        event = self.metadata.get("event_name")
 
         # format date for metadata
+        date = f"{self.metadata['start_datetime'].year}-{self.metadata['end_datetime'].year}"
         if self.metadata["start_datetime"].year == self.metadata["end_datetime"].year:
             date = str(self.metadata["start_datetime"].year)
```
l0b0, line 227:
```python
if something:
    foo = other
else:
    foo = default
```

vs.

```python
foo = default
if something:
    foo = other
```

isn't always clear cut:

- The second variant is shortest.
- The first variant is less computationally costly, because we never compute a value and then throw it away.
- The first variant is (IMO) clearer because the values are at the same indentation level.
- The first variant avoids variable re-use, which makes debugging easier.
- All that said, if `default` has a side-effect the two variants aren't equivalent.

On balance, I'd consider the advantages of the first variant to usually outweigh the shorter second variant, as long as the side-effect isn't relevant.

scripts/stac/imagery/collection.py
```diff
@@ -210,94 +210,70 @@ def write_to(self, destination: str) -> None:
...
             RURAL_AERIAL_PHOTOS,
         ]:
             return " ".join(
-                f"{name} {self.metadata['gsd']} {event or ''} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
+                f"{imagery_name} {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
             )
         if self.metadata["category"] in [DEM, DSM]:
             return " ".join(
-                f"{name} {self._elevation_title_event(event) or ''} LiDAR {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
+                f"{region} {elevation_description or ''} LiDAR {self.metadata['gsd']} {DATA_CATEGORIES[self.metadata['category']]} ({date}) {preview or ''}".split()  # pylint: disable=line-too-long
```
l0b0, line 257: One way to avoid really long lines and the `split`/`join` method would be to use something like this:

```python
" ".join(
    value
    for value in [
        region,
        elevation_description or None,
        "LiDAR",
        …]
    if value is not None
)

```

scripts/stac/tests/generate_title_test.py
l0b0: Sounds good, although I'd probably use an `else` for the fallback.

scripts/stac/imagery/collection.py
```diff
@@ -308,17 +284,13 @@ def _description(self) -> str:
...
-            desc = " ".join(
-                f"Digital Surface Model within the {region} {geographic_description or ''} region in {date}".split()
-            )
+            desc = f"Digital Surface Model within the {region} region in {date}"
         else:
             raise SubtypeParameterError(self.metadata["category"])
 
-        if event:
-            desc = desc + f", published as a record of the {event} event"
+        if self.metadata.get("event_name"):
```
paulfouquet, line 293: You could use:
```python
if event_name := self.metadata.get("event_name"):
   desc = desc + f", published as a record of the {event_name} event"
```

to avoid lookup twice. [Example](https://github.com/linz/topo-imagery/blob/dea3189b9afbb198e02888ed6ef9ebb17687a000/scripts/stac/imagery/collection.py#L56).

scripts/stac/tests/collection_test.py
```diff
@@ -24,24 +24,24 @@ def setup() -> Generator[CollectionMetadata, None, None]:
         "start_datetime": datetime(2022, 2, 2),
         "end_datetime": datetime(2022, 2, 2),
         "lifecycle": "completed",
-        "event_name": "Forest assessment",
+        "event_name": "Forest Assessment",
         "historic_survey_number": None,
-        "geographic_description": "Auckland North",
+        "geographic_description": "Auckland North Forest Assessment",
```
paulfouquet, line 29: Does the user need to duplicate this information from `event_name` to `geographic_description`?

scripts/stac/tests/generate_title_test.py
dwsilk: I think we have to keep it to whatever geographic_description is added because I can think of scenarios where this would be worse.

The event name is clearly Cyclone Gabrielle but you might receive two different datasets within the same region and want to name one Hawke's Bay - Northern Rivers Cyclone Gabrielle and one Hawke's Bay - Southern Rivers Cyclone Gabrielle. You can't just use region and event name to distinguish those. But you can keep our STAC event field as Cyclone Gabrielle and then use "Northern Rivers Cyclone Gabrielle" and "Southern Rivers Cyclone Gabrielle" as the geographic_description.

The geographic_description will be a bit of a junk field on the STAC side of things because it will contain all sorts, but that's just what I feel is required in order to be able to name things in the most helpful way.

l0b0 (approved):

scripts/stac/tests/collection_test.py
MDavidson17: Sorry that was confusing I think I did a better job explaining it here: https://github.com/linz/argo-tasks/pull/845

paulfouquet (approved):

PR #860, opened 2024-02-15
feat: save a copy of capture-area.geojson as an artifact for Argo Workflows TDE-1045

paulfouquet:
#### Motivation

In order to allow Argo Workflows UI to expose a file to the user, the file has to be stored in the local machine. The workflows can then pulled the file as an artifact output. https://github.com/linz/topo-workflows/pull/443

#### Modification

Save a copy of  `capture-area.geojson` file in a different path than the one linked to the STAC Collection.

Another option would be to only save the file locally and let Argo Workflow to save it to the target directory. I haven't choose that option to monitor AWS issues from the application rather than from Argo.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -83,7 +83,7 @@ class ImageryCollection:
 
         self.add_providers(providers)
 
-    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str) -> None:
+    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str, artifact_target: str = "tmp/") -> None:
         """Add the capture area of the Collection.
         The `href` or path of the capture-area.geojson is always set as the relative `./capture-area.geojson`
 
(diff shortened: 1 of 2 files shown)


MDavidson17 (dismissed):

scripts/stac/imagery/collection.py
```diff
@@ -83,13 +83,15 @@ def __init__(
...
 
-    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str) -> None:
+    def add_capture_area(self, polygons: List[shapely.geometry.shape], target: str, artifact_target: str = "/tmp") -> None:
         """Add the capture area of the Collection.
         The `href` or path of the capture-area.geojson is always set as the relative `./capture-area.geojson`
 
         Args:
             polygons: list of geometries
-            target: path of the capture-area-geojson file
+            target: path where the capture-area.geojson file is saved
```
l0b0, line 92: "is saved" might be confusing, as it could indicate that it should already be there when the user runs this function. It might be useful to write it from the perspective of before running the function, for example, "path where the capture-area.geojson file will be saved" or "path of the new capture-area.geojson file".

"path" also could be confusing, since the target could also be an S3 URL. Maybe "location" would be a better word?

l0b0 (approved):

MDavidson17 (approved):

PR #862, opened 2024-02-19
fix: always ignore existing overviews TDE-1046

amfage:
#### Motivation

Existing overviews were only being ignored when webp overview compression was used. This was not used by the dem_lerc preset so overviews were not being ignored for LERC compressed DEMs/DSMs and therefore when restandardising already standardised datasets the quality of the overviews could be degraded.

#### Modification

Move the `overviews=ignore_existing` `gdal_translate` configuration option to the `BASE_COG` configuration so it is always set.

#### Checklist

- [X] Tests updated
- [X] Docs updated
- [X] Issue linked in Title

### DIFF
--- a/scripts/gdal/gdal_preset.py
+++ b/scripts/gdal/gdal_preset.py
@@ -31,7 +31,7 @@ BASE_COG = [
     # An error will be raise by GDAL if it fails creating a tiff > 4GB in size
     "-co",
     "bigtiff=no",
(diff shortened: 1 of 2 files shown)


l0b0 (dismissed):

paulfouquet (commented):
Should we briefly explain why we ignore them?

paulfouquet (approved):

PR #891, opened 2024-03-05
feat: simplify footprint geometry tde-1080

paulfouquet:
#### Motivation

Since [ `max_points unlimited` has been introduced in the `gdal_footprint`](https://github.com/linz/topo-imagery/pull/872) call, generating the capture-area might take a very long time due to the complexity of the single footprint polygon that are merged.

#### Modification

Simplify single footprint geometry using the same buffer distance its used for the capture-area geometry.

From one test standardising [Canterbury - Banks Peninsula LiDAR 1m DEM (2023)](https://github.com/linz/elevation/blob/master/stac/canterbury/banks-peninsula_2023/dem_1m/2193/collection.json) dataset, the create-collection step running time:
- without simplifying footprints: 2m46s
- simplifying footprints: 12s

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/scripts/stac/imagery/capture_area.py
+++ b/scripts/stac/imagery/capture_area.py
@@ -10,6 +10,36 @@ Degree precision of ~1m (decimal places 5, https://en.wikipedia.org/wiki/Decimal
 """
 
 
+def gsd_to_float(gsd: str) -> float:
+    """Transform the gsd from its string/human readable format to a float.
+
+    Args:
+        gsd: gsd as it's passed to the cli
+
+    Returns:
+        gsd as a float
+
+    Examples:
+        >>> gsd_to_float("0.2m")
+        0.2
+    """
+    return float(gsd.replace("m", ""))
+
+
+def get_buffer_distance(gsd: float) -> float:
+    """The `gsd` (in meters) is multiplied by 2 and then by the 1m degree of precision.
+    A `buffer factor` of 2 was decided on after experimenting with different outputs,
+    details of this can be found in TDE-1049.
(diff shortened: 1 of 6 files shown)


scripts/stac/imagery/capture_area.py
```diff
@@ -10,6 +10,36 @@
 """
 
 
+def gsd_to_float(gsd: str) -> float:
```
l0b0, line 13: Could we require the parameter to be a decimal number (in metres), so that we don't have to deal with the unit? We don't support any other units, after all.

paulfouquet: I have to check how it is used by the other tasks (if they need the unit, like path generation in argo-task). Ideally we would like the user to pass only one parameter for the GSD.

paulfouquet: I have checked and it is actually not used by any other tasks (except create-collection). 
The [gsd is extracted from the TIFF](https://github.com/linz/argo-tasks/blob/0c3e4a60b7ec8044accda9050199363b876e9e0f/src/commands/path/path.generate.ts#L173) in argo-tasks, which makes me thinking that we could potentially do the same here, rather than passing it as a parameter.

paulfouquet: This should be dealt in another ticket/PR -> TDE-1085

l0b0 (approved):

amfage (approved):

PR #893, opened 2024-03-05
fix: Treat "no data" value as a decimal number TDE-1077

l0b0:
#### Modification

It's represented as a decimal number string in the `gdalinfo -json`
output, so we should treat it as such in the code. This should detect
any values not numerically equal to -9999.

#### Checklist

- [x] Tests updated
- [ ] Docs updated (N/A)
- [x] Issue linked in Title

### DIFF
--- a/scripts/files/file_tiff.py
+++ b/scripts/files/file_tiff.py
@@ -1,10 +1,13 @@
 import json
+from decimal import Decimal
 from enum import Enum
 from typing import Any, Dict, List, Optional
 from urllib.parse import unquote
 
 from scripts.gdal.gdal_helper import GDALExecutionException, gdal_info, run_gdal
 from scripts.gdal.gdalinfo import GdalInfo
+
+DEFAULT_NO_DATA_VALUE = Decimal(-9999)
 
 
 class FileTiffErrorType(str, Enum):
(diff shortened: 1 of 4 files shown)


paulfouquet (approved):

scripts/files/file_tiff.py
```diff
@@ -1,11 +1,14 @@
 import json
+from decimal import Decimal
 from enum import Enum
 from typing import Any, Dict, List, Optional
 from urllib.parse import unquote
 
 from scripts.gdal.gdal_helper import GDALExecutionException, gdal_info, run_gdal
 from scripts.gdal.gdalinfo import GdalInfo
 
+DEFAULT_NO_DATA_VALUE = Decimal(-9999)
```
paulfouquet, line 10: Not part of this PR but if you want you could add a comment about why we are using `-9999`

l0b0: Can do! Do you have a reference? I think it was mentioned recently, but I can't find it in Slack.

l0b0: #901 is in draft, because it depends on this PR.

blacha (approved):

PR #894, opened 2024-03-05
build: Bump Python version TDE-1077

l0b0:
#### Motivation

Use a strict version string to make sure we're running with the same
version across the board.

#### Modification

Use Python 3.10.12.

#### Checklist

- [ ] Tests updated (N/A)
- [ ] Docs updated (N/A)
- [x] Issue linked in Title


.github/workflows/format-tests.yml
```diff
@@ -11,10 +11,9 @@ jobs:
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
 
-      - name: Use Python "3.10.6"
-        uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v4
+      - uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v4
         with:
-          python-version: "3.10.6"
+          python-version: "3.10.13"
```
paulfouquet, line 16: Would it be better to use the same version as we use locally, even if it's to hold it back here?

l0b0: The versions we use locally isn't going to be the same across all our machines. I'd rather there were no differences between the different Python interpreters we use in CI. I've set the version to a specific one to enforce that, which achieves that. We should probably be using `pyenv` or something similar to make sure we're using the right version locally, rather than relying on the OS version.

l0b0:
Not necessary.

PR #906, opened 2024-03-11
fix: wrong vsis3 path logged

paulfouquet:
#### Motivation

The `vsis3` path is logged in order to facilitate the user to investigate an issue on the output tile by running gdal directly with the path copied from logs.
The path is currently logged wrongly: when the script is ran on Argo Workflows, the standardised path is already a S3 path (as target is S3 scratch bucket).
The result with the old implementation gave a path like `/vsis3/linz-workflows-scratch/2024-02/14-test-northland-2018-2020-dsm-1mtfvvz/test-northland-2018-2020-dsm-1mtfvvz-standardise-validate-1780775412/s3:/linz-workflows-scratch/2024-02/14-test-northland-2018-2020-dsm-1mtfvvz/flat/AX27_10000_0105.tiff`

#### Modification

Change the implementation to use `get_vfs_path` which replaces `s3://` by `/vsis3/`.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/scripts/gdal/gdal_helper.py
+++ b/scripts/gdal/gdal_helper.py
@@ -34,6 +34,9 @@ def get_vfs_path(path: str) -> str:
 
     Returns:
         str: the path modified to comply with the corresponding storage service.
+    Examples:
+        >>> get_vfs_path("s3://scratch-bucket/path/to/file.tiff")
+        '/vsis3/scratch-bucket/path/to/file.tiff'
     """
     return path.replace("s3://", "/vsis3/")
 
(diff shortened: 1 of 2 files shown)


blacha (approved):
unrelated, but how well does this path logic work in a windows environment? I am assuming we have just out of scopped windows though.

l0b0 (commented):
The code LGTM, but is the example in the PR description right? There's a "s3:" in there.

paulfouquet:
> The code LGTM, but is the example in the PR description right? There's a "s3:" in there.

Yes that was the issue

paulfouquet:
> unrelated, but how well does this path logic work in a windows environment? I am assuming we have just out of scopped windows though.

This path logic is triggered only when using Argo, which has a target path (where standardised files are saved) of a S3 bucket. 
If the path is a local path, we don't modify it

l0b0 (approved):

PR #964, opened 2024-05-15
fix: add draft lifecycle tag TDE-1161

MDavidson17:
#### Motivation

Align code with title naming convention documentation for [imagery](https://github.com/linz/imagery/blob/master/docs/naming.md#lifecycle) and [elevation](https://github.com/linz/elevation/blob/master/docs/naming.md#lifecycle)

#### Modification

Add `- Draft` to the end of a dataset title if the lifecycle status is `ongoing`

#### Checklist

- [x] Tests updated
- [ ] Docs updated - Docs are correct updating code to match
- [x] Issue linked in Title

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -265,18 +265,20 @@ class ImageryCollection:
             imagery_name = region
             elevation_description = None
 
-        # determine if dataset is preview
+        # determine if the dataset title requires a lifecycle tag
         if self.metadata.get("lifecycle") == "preview":
-            preview = "- Preview"
+            lifecycle_tag = "- Preview"
+        elif self.metadata.get("lifecycle") == "ongoing":
+            lifecycle_tag = "- Draft"
         else:
-            preview = None
+            lifecycle_tag = None
 
         if self.metadata["category"] == SCANNED_AERIAL_PHOTOS:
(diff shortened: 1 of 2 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -265,18 +265,20 @@ def _title(self) -> str:
...
-            preview = None
+            lifecycle_tag = None
 
         if self.metadata["category"] == SCANNED_AERIAL_PHOTOS:
             if not historic_survey_number:
                 raise MissingMetadataError("historic_survey_number")
             return " ".join(
                 value
-                for value in [imagery_name, self.metadata["gsd"], historic_survey_number, f"({date})", preview or None]
+                for value in [imagery_name, self.metadata["gsd"], historic_survey_number, f"({date})", lifecycle_tag or None]
```
l0b0, line 281: Is the `or None` part necessary here? That would only really be applicable if the value is a falsy value other than `None`.

l0b0 (approved):

scripts/stac/imagery/collection.py
MDavidson17: I've removed the `or None` steps I believe that has been carried over from when the values were falsy but it has since been updated so is no longer required.

paulfouquet (approved):

PR #971, opened 2024-05-19
fix: Make build compatible with new GDAL container TDE-1179

l0b0:
#### Motivation

Make build compatible with [GDAL 3.9 container](https://github.com/linz/topo-imagery/pull/962).

#### Modification

- Use multi-stage build to avoid including build tools in the final image.
- Install Poetry using pipx, as [recommended](https://python-poetry.org/docs/#installation).
- Use a different working directory in the first stage, to avoid the possibility of confusing the two or unintended interactions between them.
- Use a virtualenv to avoid clobbering OS packages.

#### Checklist

- [ ] Tests updated (N/A)
- [ ] Docs updated (N/A)
- [x] Issue linked in Title

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+set -o errexit
+
+. /venv/bin/activate
+
+exec "$@"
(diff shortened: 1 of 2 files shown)


scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -o errexit -o xtrace
```
amfage, line 3: Do we need `xtrace`?

l0b0: D'oh! Fixed.

amfage (approved):

blacha (approved):

PR #973, opened 2024-05-21
feat: add checksum metadata to s3 write function TDE-1181

MDavidson17:
#### Motivation

Store checksum as s3 metadata so that we can determine if files have changed

#### Modification

- add checksum to s3 object [metadata](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/object/put.html#:~:text=S3%20on%20Outposts.-,Metadata%20(dict)%20%E2%80%93,(string)%20%E2%80%93,-ServerSideEncryption%20(string)) as `filechecksum`

#### Output
image
nb: in the code, it is defined as `filechecksum` and in the tests you use `resp["Metadata"]["filechecksum"]` but something else adds the prefix `x-amz-meta-` here that I can't remove.

#### Checklist

- [x] Tests updated - has also been tested manually
- [ ] Docs updated - no docs
- [x] Issue linked in Title

### DIFF
--- a/scripts/files/fs_s3.py
+++ b/scripts/files/fs_s3.py
@@ -9,6 +9,7 @@ from linz_logger import get_log
 
 from scripts.aws.aws_helper import get_session, parse_path
 from scripts.logging.time_helper import time_in_ms
+from scripts.stac.util import checksum
 
 if TYPE_CHECKING:
     from mypy_boto3_s3 import S3Client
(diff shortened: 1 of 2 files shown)


blacha:
the `x-amz-meta-` is a prefix forced upon any non standard tag https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata

do we have a preference on the name, im assuming `:` is not allowed, do we want to just shorten it to `checksum`, checksum is very generic though, and doesnt tell you what it is, having something like `multihash` might be better?

scripts/files/fs_s3.py
```diff
@@ -32,13 +33,14 @@ def write(destination: str, source: bytes, content_type: Optional[str] = None) -
     s3_path = parse_path(destination)
     key = s3_path.key
     s3 = resource("s3")
+    fileChecksum = checksum.multihash_as_hex(source)
 
     try:
         s3_object = s3.Object(s3_path.bucket, key)
         if content_type:
-            s3_object.put(Body=source, ContentType=content_type)
+            s3_object.put(Body=source, ContentType=content_type, Metadata={"filechecksum": fileChecksum})
```
l0b0, line 41: It might be useful to call this something like "multihash" to make it more self-documenting. "file" is probably redundant in the metadata name.

paulfouquet:
> do we have a preference on the name, im assuming `:` is not allowed, do we want to just shorten it to `checksum`, checksum is very generic though, and doesnt tell you what it is, having something like `multihash` might be better?

:+1: `multihash`

l0b0 (approved):

blacha (approved):

PR #974, opened 2024-05-23
fix: should use unique file names when copying files to standardise TDE-1186

paulfouquet:
Co-authored-by: Alice Fage <afage@linz.govt.nz>

﻿#### Motivation

When a source input contains more than one file with the same name, standardising script should not overwrite them at the copy but copy them all.

#### Modification

Generate a filename based on the multihash of the input file original path

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


scripts/translate_ascii.py
```diff
@@ -51,7 +51,7 @@ def main() -> None:
         p.close()
         p.join()
 
-        write_all(inputs=tiffs, target=arguments.target)
+        write_all(inputs=tiffs, target=arguments.target, generated_name=False)
```
l0b0, line 54: Why don't we want to use the generated name here?

scripts/files/fs.py
```diff
@@ -141,20 +142,27 @@ def write_sidecars(inputs: List[str], target: str, concurrency: Optional[int] =
...
 
-def write_file(executor: ThreadPoolExecutor, input_: str, target: str) -> Future[str]:
+def write_file(executor: ThreadPoolExecutor, input_: str, target: str, generated_name: Optional[bool] = True) -> Future[str]:
     """Read a file from a path and write it to a target path.
-
     Args:
         executor: A ThreadPoolExecutor instance.
         input_: A path to a file to read.
         target: A path to write the file to.
+        generated_name: create a target file name based on multihash the source filename
```
blacha, line 151: generated, implies the name was already generated, how about `generate_name` ?

scripts/translate_ascii.py
paulfouquet: Because the files that are being copied need to have their original name as they are not temporary files.

l0b0 (approved):

blacha (approved):

PR #998, opened 2024-07-08
fix: concurrency file order TDE-1213

amfage:
#### Motivation

When supplying multiple source locations to use for standardising, if there are overlapping source images, the VRT created does not honour the priority order for the supplied datasets. This Pull Request is to retain the priority of the images for merging.

#### Modification

Move away from using `as_completed` for parallel processing so that the output file list is in the same order as the input file list.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/scripts/files/fs.py
+++ b/scripts/files/fs.py
@@ -141,7 +141,7 @@ def write_sidecars(inputs: list[str], target: str, concurrency: int | None = 4) 
     for future in results:
         future_ex = future.exception()
         if isinstance(future_ex, NoSuchFileError):
-            get_log().info("No sidecar file found; skipping", error="future.exception()")
+            get_log().info("No sidecar file found; skipping", error=future.exception())
         else:
             get_log().info("wrote_sidecar_file", path=future.result())
 
--- a/scripts/files/fs.py
+++ b/scripts/files/fs.py
@@ -1,5 +1,5 @@
 import os
-from concurrent.futures import Future, ThreadPoolExecutor, as_completed
+from concurrent.futures import Future, ThreadPoolExecutor
 from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


scripts/files/tests/fs_test.py
```diff
@@ -89,6 +89,18 @@ def test_write_sidecars_one_found(capsys: CaptureFixture[str], subtests: SubTest
     rmtree(target)
 
 
+def test_write_all_in_order(setup: str) -> None:
```
blacha, line 92: does this test fail before the change?

scripts/files/fs.py
```diff
@@ -130,26 +133,28 @@ def write_sidecars(inputs: list[str], target: str, concurrency: int | None = 4)
...
-                get_log().info("No sidecar file found; skipping", path=future_ex.path)
-            else:
-                get_log().info("wrote_sidecar_file", path=future.result())
+        for input_ in inputs:
+            results.append(executor.submit(write_file, input_, target))
+
+    for future in results:
+        future_ex = future.exception()
+        if isinstance(future_ex, NoSuchFileError):
+            get_log().info("No sidecar file found; skipping", error="future.exception()")
```
blacha, line 144: why is the error in quotes?

scripts/files/tests/fs_test.py
amfage: I have updated the test as discussed to one that fails before the code change.
Example of `written_files` list before the change:
`['/tmp/tmpjdifie0u/5', '/tmp/tmpjdifie0u/0', '/tmp/tmpjdifie0u/3', '/tmp/tmpjdifie0u/1', '/tmp/tmpjdifie0u/7', '/tmp/tmpjdifie0u/2', '/tmp/tmpjdifie0u/9', '/tmp/tmpjdifie0u/4', '/tmp/tmpjdifie0u/6', '/tmp/tmpjdifie0u/8']`
and after:
`['/tmp/tmp7gjme7sj/0', '/tmp/tmp7gjme7sj/1', '/tmp/tmp7gjme7sj/2', '/tmp/tmp7gjme7sj/3', '/tmp/tmp7gjme7sj/4', '/tmp/tmp7gjme7sj/5', '/tmp/tmp7gjme7sj/6', '/tmp/tmp7gjme7sj/7', '/tmp/tmp7gjme7sj/8', '/tmp/tmp7gjme7sj/9']`

scripts/files/tests/fs_test.py
```diff
@@ -89,6 +89,21 @@ def test_write_sidecars_one_found(capsys: CaptureFixture[str], subtests: SubTest
...
+    inputs: list[str] = []
+    i = 0
+    while i < 10:
+        path = Path(os.path.join(setup, str(i)))
+        if i % 2 == 0:
+            path.write_text("a" * 1000 * 1000, encoding="utf-8")  # 1MB
```
blacha, line 98: NIT: would be good to generating a bunch of garbage once rather than generating it every time

blacha (approved):

blacha (approved):

dwsilk (approved):

PR #1012, opened 2024-07-28
test: init automatic tests for collection_from_items.py script TDE-1227

paulfouquet:
### Motivation

Having automatic tests for `collection_from_items.py` script.

### Modifications

- Refactor `collection_from_items.py` `main()` function so it can be easily called passing a list of arguments.
- Add a simple test that verifies the `collection.json` file is created on a usual use case.

### Verification

run `pytest`

### DIFF
--- a/scripts/collection_from_items.py
+++ b/scripts/collection_from_items.py
@@ -93,7 +93,6 @@ def parse_args(args: List[str] | None) -> Namespace:
 def main(args: List[str] | None = None) -> None:
     arguments = parse_args(args)
     uri = arguments.uri
-    print(arguments.start_date)
 
     providers: list[Provider] = []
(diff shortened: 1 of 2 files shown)


scripts/collection_from_items.py
```diff
@@ -85,8 +86,14 @@ def main() -> None:
...
 
-    arguments = parser.parse_args()
+    return parser.parse_args(args)
+
+
+# pylint: disable=too-many-locals
+def main(args: List[str] | None = None) -> None:
+    arguments = parse_args(args)
     uri = arguments.uri
+    print(arguments.start_date)
```
blacha, line 96:
```suggestion
    print(arguments.start_date)
```

blacha (approved):

PR #1038, opened 2024-08-29
ci: publish container to gh on pr label container TDE-1252

paulfouquet:
### Motivation

It is useful to be able to publish a container from a PR code base so we can test the container using external environment such as Kubernetes.

### Modifications

- GH Actions workflow `containers.yml` will run when a `container` label is added to a PR. The container will only be published on the GH package repository and be tagged `pr-[PR-NUMBER]`

### Verification

You can test it in this PR.
Also tested on a fork with the full process of merging the PR for non regression.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -5,10 +5,12 @@ on:
     branches: ["master"]
     types:
       - completed
+  pull_request:
+    types: labeled
 jobs:
   build-containers:
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event.label.name == 'container' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/containers.yml
```diff
@@ -68,8 +70,12 @@ jobs:
           result-encoding: string
           script: |
             const tags = [];
-            tags.push('ghcr.io/${{ github.repository }}:latest');
-            tags.push('ghcr.io/${{ github.repository }}:${{ steps.version.outputs.version }}');
+            if ('${{github.event.label.name}}' == 'container') {
+              tags.push('ghcr.io/${{ github.repository }}:pr-${{github.event.number}}');
+            } else{
```
amfage, line 75:
```suggestion
            } else {
```

amfage (approved):

PR #1043, opened 2024-09-02
feat: add derived_from links to item TDE-1251

paulfouquet:
### Motivation

Some tiles could be generated from a combination of existing tiles (example: `AV29.tiff` - 50k is made of the combination of `northland/northland_2018-2020/dem_1m/2193/AV29_10000_0505.tiff`,  `northland/northland_2018-2020/dem_1m/2193/AV29_10000_0504.tiff`, etc.). Storing this information in the associated STAC Item of the tile allows:
- the data consumer to be aware of which data and version has been used to make this tile
- the system to detect if any of the source data has changed since its generation (comparing stored checksum vs actual checksum of the `derived_from` links)

### Modifications

- `standardise_validate.py` script accepts a new attribute `include_derived` in its `--from-file` input format.
- if `include_derived` is `True`, the created STAC Item will add the `derived_from` links based on the `input` files used to create the `output` tile.

### Verification

Few workflows have been run and the container is available in GH Packages.


amfage (dismissed):

scripts/files/file_tiff.py
```diff
@@ -39,6 +41,12 @@ def __init__(
             paths_original.append(unquote(p))
 
         self._paths_original = paths_original
+        self._derived_from_paths = None
+        if include_derived:
+            # Transform the TIFF paths to JSON path to point to STAC Items,
+            # assuming the STAC Items are in the same directory as the TIFF files
```
l0b0, line 47: Multi-line comments should usually be in `"""` blocks, to keep the lines semantically together.

scripts/stac/imagery/tests/create_stac_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_create_item_with_derived_from(tmp_path: Path) -> None:
+    derived_from_path = tmp_path / "derived_from_item.json"
+    derived_from_path.write_text('{"type": "Feature", "id": "fake_item"}')
+    fake_gdal_info: GdalInfo = cast(
+        GdalInfo, {"wgs84Extent": {"type": "Polygon", "coordinates": [[[0, 1], [1, 1], [1, 0], [0, 0]]]}}
+    )
+
+    item = create_item(
+        "./scripts/tests/data/empty.tiff", "2024-01-01", "2024-01-02", "abc123", fake_gdal_info, [f"{derived_from_path}"]
```
l0b0, line 16: `f"{derived_from_path}"` can be simplified to `derived_from_path`; see also other instances of the same.

scripts/files/file_tiff.py
paulfouquet: `"""` is usually for [docstring](https://peps.python.org/pep-0257/) which I don't want to add here

l0b0 (approved):

PR #1052, opened 2024-09-09
feat: determine start and end datetimes when item include derived TDE-1258

paulfouquet:
### Motivation

When a STAC Item is derived from other STAC Items, its `properties.start_datetime` and `properties.end_datetime` should be determined from the derived  items.

### Modifications

- list all the `start_datetime` and `end_datetime` of derived Items to determine the new Item start and end datetime.

### Verification

tests

### DIFF
--- a/scripts/stac/imagery/create_stac.py
+++ b/scripts/stac/imagery/create_stac.py
@@ -2,7 +2,7 @@ import json
 
 from linz_logger import get_log
 
-from scripts.datetimes import parse_rfc_3339_datetime, utc_now
+from scripts.datetimes import utc_now
 from scripts.files.files_helper import get_file_name_from_path
 from scripts.files.fs import read
 from scripts.files.geotiff import get_extents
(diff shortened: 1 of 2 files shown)


scripts/stac/imagery/create_stac.py
```diff
@@ -40,15 +42,24 @@ def create_item(
...
-    item.update_spatial(geometry, bbox)
-    item.add_collection(collection_id)
 
     if derived_from is not None:
+        start_date_list = []
+        end_date_list = []
         for derived in derived_from:
+            derived_item_content = read(derived)
+            derived_stac = json.loads(derived_item_content.decode("UTF-8"))
+            start_date_list.append(derived_stac["properties"]["start_datetime"])
```
l0b0, line 52: Since we only need one of each of these, rather than adding all of them to a list and then computing the min/max, we can continuously replace the value if it's lower than the existing value:

```suggestion
        start_datetime = min(start_datetime, derived_stac["properties"]["start_datetime"])
```

You can compare RFC 3339 datetime strings without converting them to datetimes, so this should even be faster.

l0b0 (approved):

PR #1060, opened 2024-09-10
refactor: Pull out enum for presets

l0b0:
### Motivation

Reduce the amount of magic strings in the code base

### Verification

None.


paulfouquet (approved):
Seems reasonable to me since we've got end-to-end tests to cover user inputs

amfage (approved):

PR #1061, opened 2024-09-10
refactor: Use numeric enum for EPSG codes

l0b0:
### Motivation

Avoid hardcoded values.

### Verification

None.

### DIFF
--- a/scripts/gdal/gdal_helper.py
+++ b/scripts/gdal/gdal_helper.py
@@ -19,11 +19,11 @@ class GDALExecutionException(Exception):
     pass
 
 
-class EpsgCode(str, Enum):
-    EPSG_2193 = "EPSG:2193"
-    """ NZGD2000 / New Zealand Transverse Mercator 2000 (NZTM) """
-    EPSG_4326 = "EPSG:4326"
-    """ WGS84 - World Geodetic System 1984"""
+class EpsgNumber(int, Enum):
+    EPSG_NZGD2000 = 2193
+    """New Zealand Transverse Mercator 2000 (NZTM)"""
+    EPSG_WGS_1984 = 4326
+    """World Geodetic System 1984"""
 
 
 def get_vfs_path(path: str) -> str:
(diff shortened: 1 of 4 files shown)


scripts/standardising.py
```diff
@@ -220,7 +220,7 @@ def standardising(
                         [
                             "gdal_footprint",
                             "-t_srs",
-                            EpsgCode.EPSG_4326,
```
l0b0, line 223: I don't understand how this could've worked before, because:

```
>>> EpsgCode.EPSG_4326
<EpsgCode.EPSG_4326: 'EPSG:4326'>
>>> str(EpsgCode.EPSG_4326)
'EpsgCode.EPSG_4326'
```

paulfouquet: Something to deal with `subprocess.run()` I guess?

```
>>> command: list[str]= ["echo", EpsgCode.EPSG_4326]
>>> proc = subprocess.run(command, stdout=subprocess.PIPE, check=True)
>>> print(proc.stdout.decode().strip())
EPSG:4326
```

paulfouquet (approved):

scripts/gdal/gdal_helper.py
```diff
@@ -19,11 +19,11 @@ class GDALExecutionException(Exception):
     pass
 
 
-class EpsgCode(str, Enum):
-    EPSG_2193 = "EPSG:2193"
-    """ NZGD2000 / New Zealand Transverse Mercator 2000 (NZTM) """
-    EPSG_4326 = "EPSG:4326"
-    """ WGS84 - World Geodetic System 1984"""
+class EpsgNumber(int, Enum):
+    EPSG_NZGD2000 = 2193
```
blacha, line 23: NZGD2000 is [EPSG:4167](https://epsg.io/4167)

l0b0: I was just basing this on the original comment. Would `NZTM` be a better name?

l0b0: See #1063.

PR #1065, opened 2024-09-11
feat: make start_datetime and end_datetime optional when includeDerived is True TDE-1258

paulfouquet:
### Motivation

`start_datetime` and `end_datetime` should be optional in the case of the output tiles include derived items.

### Modifications

- make the input parameters not always required

### Verification

Ran a couple of test workflows with the following:
- empty datetimes
- bad format datetime
- good format datetime

### DIFF
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -47,10 +47,16 @@ def main() -> None:
     parser.add_argument("--cutline", dest="cutline", help="Optional cutline to cut imagery to", required=False, nargs="?")
     parser.add_argument("--collection-id", dest="collection_id", help="Unique id for collection", required=True)
     parser.add_argument(
-        "--start-datetime", dest="start_datetime", help="Start datetime in format YYYY-MM-DD", type=valid_date, required=True
+        "--start-datetime",
+        dest="start_datetime",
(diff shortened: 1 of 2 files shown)


scripts/standardise_validate.py
```diff
@@ -47,10 +47,16 @@ def main() -> None:
     parser.add_argument("--cutline", dest="cutline", help="Optional cutline to cut imagery to", required=False, nargs="?")
     parser.add_argument("--collection-id", dest="collection_id", help="Unique id for collection", required=True)
     parser.add_argument(
-        "--start-datetime", dest="start_datetime", help="Start datetime in format YYYY-MM-DD", type=valid_date, required=True
+        "--start-datetime",
+        dest="start_datetime",
+        help="Start datetime in format YYYY-MM-DD. Only optional if includeDerived.",
+        type=valid_date,
```
l0b0, line 53: Why not make this `required=False` instead of allowing `--start-datetime=`?

paulfouquet: Is the default value not `False`?

l0b0 (approved):

scripts/standardise_validate.py
paulfouquet: Sorry, I did forgot to revert that before PR

PR #1079, opened 2024-09-18
fix: start_datetime and end_datetime must be init if derived_from TDE-1258

paulfouquet:
### Motivation

Test was incorrect.

### Modifications

- update test
- init start_datetime and end_datetime if derived_from

### Verification

run pytest and a local test

### DIFF
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -24,6 +24,8 @@ def str_to_bool(value: str) -> bool:
 
 def main() -> None:
     # pylint: disable-msg=too-many-locals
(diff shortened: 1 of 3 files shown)


scripts/standardise_validate.py
```diff
@@ -24,6 +24,8 @@ def str_to_bool(value: str) -> bool:
 
 def main() -> None:
     # pylint: disable-msg=too-many-locals
+    # pylint: disable-msg=too-many-branches
```
paulfouquet, line 27: see TDE-1270

l0b0 (approved):

PR #1080, opened 2024-09-19
feat: Add topo-imagery version information to STAC TDE-1265

l0b0:
### Motivation

As a data maintainer, geospatial data engineer or data consumer, I would like to reference the technical configuration used to create the TIFFs in our ODR datasets in order to be able to recreate the TIFFs or carry out troubleshooting.

### Modifications

Add `processing:software`, `processing:version`, and `processing:datetime` item properties.

### Verification

`pytest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The scripts have been implemented to be run inside the Docker container only. Th
 - Build the `Docker` image:
 
 ```bash
-docker build . -t topo-imagery
+docker build --build-arg=GIT_HASH=dev --build-arg=GIT_VERSION=dev --tag=topo-imagery .
 ```
 
 - Running `standardising_validate.py` script
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The scripts have been implemented to be run inside the Docker container only. Th
 - Build the `Docker` image:
 
 ```bash
-docker build --build-arg=GIT_HASH=dev --build-arg=GIT_VERSION=dev --tag=topo-imagery .
+docker build --tag=topo-imagery .
 ```
 
 - Running `standardising_validate.py` script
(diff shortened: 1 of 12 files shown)


scripts/standardising.py
```diff
@@ -50,11 +51,11 @@ def run_standardising(
 
     Returns:
         a list of FileTiff wrapper
+        :param gdal_version:
```
paulfouquet, line 54: Should be in the Args list?

README.md
```diff
@@ -27,7 +27,7 @@ The scripts have been implemented to be run inside the Docker container only. Th
 - Build the `Docker` image:
 
 ```bash
-docker build . -t topo-imagery
+docker build --build-arg=GIT_HASH=dev --build-arg=GIT_VERSION=dev --tag=topo-imagery .
```
blacha, line 30: do we need to make the local build commands more complicated? what value does adding "dev"/"dev" add here?

scripts/standardising.py
amfage: Oh good catch, how strange. IDEs gone rogue I think!

README.md
amfage: There was some discussion about the best way to check that these exist when running in CI, open to suggestions! I think you are right we should change it as I can foresee they will end up commented out in the Dockerfile which would be a bad sign.

blacha: This is the first command in the README.md it is likely what people will run to first build this system, is making this command much more complicated to add "dev" "dev" seem excessive?

l0b0: The reason for making these mandatory was to ensure that we don't forget to add them, which would only be detected at runtime.

scripts/standardising.py
l0b0: Apropos - what *is* the format we're using here called? Then I can probably configure IDEA to use that.

README.md
l0b0: Changed to make these optional.

blacha (dismissed):

blacha (approved):

PR #1086, opened 2024-09-23
fix: Use GeoJSON media type for item links TDE-1254

l0b0:
### Motivation

https://github.com/radiantearth/stac-spec/blob/master/commons/links.md#stac-media-types

### Modifications

Change item links from media type `application/json` to `application/geo+json`.

### Verification

`pytest`

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -163,7 +163,7 @@ class ImageryCollection:
                 Link(
                     path=item_self_link["href"],
                     rel=Relation.ITEM,
-                    media_type=StacMediaType.JSON,
+                    media_type=StacMediaType.GEOJSON,
                     file_content=dict_to_json_bytes(item),
                 ).stac
             )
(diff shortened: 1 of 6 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -163,7 +163,7 @@ def add_item(self, item: dict[Any, Any]) -> None:
                 Link(
                     path=item_self_link["href"],
                     rel=Relation.ITEM,
-                    media_type=StacMediaType.JSON,
+                    media_type=StacMediaType.GEOJSON,
```
amfage, line 166: I think there may be other places this needs to change as well, such as https://github.com/linz/argo-tasks/blob/f1b559c79aa7ab5a942470354ccf1761eef6d346/src/commands/copy/copy-worker.ts#L37

l0b0: Are those changes required for this PR to go through?

amfage: After inspection the link above could be an issue but that only applies to unknown content types, which these STAC files should not be included in. Items are currently being saved to S3 and copied as `application/geo+json` .
image

amfage (approved):

paulfouquet (approved):

PR #1087, opened 2024-09-23
test: Allow minor polygon differences

l0b0:
### Motivation

Make the test pass in both GDAL 3.8.5 and 3.9.2, without reducing the accuracy of the assertion significantly.

### Modifications

Change the polygon tolerance from exactly zero to 1 part in 10 quadrillion.

### Verification

`pytest`

### DIFF
--- a/scripts/stac/imagery/tests/capture_area_test.py
+++ b/scripts/stac/imagery/tests/capture_area_test.py
@@ -1,3 +1,4 @@
+from decimal import Decimal
 from typing import cast
 
 from shapely import get_exterior_ring, is_ccw
@@ -84,9 +85,9 @@ def test_generate_capture_area_rounded() -> None:
 
     # Using GSD of 0.2m
     # Generate the capture area of the polygons that don't have a gap between each other
-    capture_area_expected = generate_capture_area(polygons_no_gap, 0.2)
+    capture_area_expected = generate_capture_area(polygons_no_gap, Decimal("0.2"))
     # Generate the capture area of the polygons that have a gap between each other
-    capture_area_result = generate_capture_area(polygon_with_gap, 0.2)
+    capture_area_result = generate_capture_area(polygon_with_gap, Decimal("0.2"))
     print(f"Polygon no gap A: {to_feature(polygons_no_gap[0])}")
     print(f"Polygon no gap B: {to_feature(polygons_no_gap[1])}")
     print(f"Polygon with gap A: {to_feature(polygon_with_gap[0])}")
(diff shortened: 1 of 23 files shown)


scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -70,7 +70,7 @@ def test_merge_polygons_with_rounding_margin_too_big() -> None:
     print(f"GeoJSON expected: {to_feature(expected_merged_polygon)}")
     print(f"GeoJSON result: {to_feature(merged_polygons)}")
 
-    assert merged_polygons.equals_exact(expected_merged_polygon, tolerance=0.0)
+    assert merged_polygons.equals_exact(expected_merged_polygon, tolerance=0.000_000_000_000_000_1)
```
blacha, line 73: where did 1e-16 come from? isn't that much less than the accuracy of a float (or are these stored as doubles?)

for our usecases on lat,lon anything less than 1e-10 is well outside our margin of error, in other places we just hard code a error margin to something like Epsilon?

l0b0: It was the smallest tolerance (I only tested multiples of 10) which made the test pass on both versions of GDAL.

blacha: I would prefer we choose a precision that is relevant to us eg 1e-10 1/100th mm or 1e-12 and store it as a rounding const somewhere, so then when we run into more floating point math rounding issues we can just reference the const.

We could also use epslion (~2.2e-16) Not sure how you get it in python in the browser you can use `Number.EPSLION`

l0b0: Epsilon is usually the smallest representable number of the relevant type, so it's pretty much guaranteed to be too small.

Are the units of these polygons always in metres? If so, something like 1e-6 (1 micrometre) should be plenty. If it's anything big, like degrees, it's going to be much more difficult to come up with a specific value which is sufficiently precise.

blacha: degrees map pretty generally to meters at a 111KM per degree level, so 8DP of degrees is 1mm at the equator and slightly less in New Zealand which is somewhere near 80KM per degree 

ref: https://en.wikipedia.org/wiki/Decimal_degrees

> Epsilon is usually the smallest representable number of the relevant type, so it's pretty much guaranteed to be too small.

1e-16 is smaller than Epsilon.

```typescript
0.000_000_000_000_000_1 < Number.EPSILON // true
```

l0b0: Looks like I misunderstood how that works. Implementing…

blacha (approved):

PR #1088, opened 2024-09-23
fix: round footprints to 8 decimal places TDE-1237

amfage:
### Motivation

As a Geospatial Data Engineer I want the geojson files my systems create to contain an appropriate precision for the data being stored so that users aren’t seeing unnecessary data.

### Modifications

Use `COORDINATE_PRECISION=8` `gdal_footprint` layer configuration option when creating the footprint files that are used to create the capture area.

### Verification

Ran manually. Examples before and after: [capture_areas.zip](https://github.com/user-attachments/files/17104010/capture_areas.zip).
Added footprint end-to-end CI test.


l0b0 (dismissed):

.github/workflows/format-tests.yml
```diff
@@ -50,6 +50,12 @@ jobs:
           docker run  -v "${{ runner.temp }}:/tmp/" topo-imagery python3 standardise_validate.py --from-file ./tests/data/aerial.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/cutline/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01 --cutline ./tests/data/cutline_aerial.fgb --gsd 10m --create-footprints=true
           cmp --silent "${{ runner.temp }}/cutline/BG35_1000_4829.tiff" ./scripts/tests/data/output/BG35_1000_4829_cut.tiff
 
+      - name: End to end test - Footprint
+        run: |
+          docker run  -v "${{ runner.temp }}:/tmp/" topo-imagery python3 standardise_validate.py --from-file ./tests/data/aerial.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01 --gsd 10m --create-footprints=true
+          jq 'select(.xy_coordinate_resolution == 1E-8) // error("Wrong or missing X/Y coordinate resolution")' "${{ runner.temp }}/BG35_1000_4829_footprint.geojson"
```
blacha, line 56: could this be the same as the other commands with a `cmp` ?

amfage: The geojson file contains the temporary file path, so it always fails.

blacha: how about removing the temporary file path then doing a cmp? with a comment explaining why

```
# file_path is a random temporary file, remove it so we can compare
cat foo.geojson | jq "del(.properties.file_path)"  > sanatized.geojson
```

l0b0: What's the advantage of using `cmp`? Using `jq` we get more information, in the form of an error message.

blacha: consistency? 

If we want to write assertions off these docker commands  they should be written inside of a proper testing framework.

l0b0: > consistency?

Consistency is not the only goal of what we do, though. The `jq` assertions are more useful (giving actual error messages) and more explicit about the things we care about.

> If we want to write assertions off these docker commands they should be written inside of a proper testing framework.

That would be great, but is outside the scope of this two-line functional change.

blacha: > Consistency is not the only goal of what we do, though. The jq assertions are more useful (giving actual error messages) and more explicit about the things we care about.

In this case the jq assertions are doing a lot less checking than what a `cmp` would check, it is only checking that their is one point, its not checking that additional properties were added.

Being consistent with a `cmp` would be preferred here.

blacha (approved):

PR #1089, opened 2024-09-23
feat!: Auto-fill GSD unit TDE-1211

l0b0:
### Motivation

This is always in metres, so we shouldn't ask the end user to fill it in.

### Modifications

Change GSD parameters to be decimal numbers, rather than strings

### Verification

`pytest`

### DIFF
--- a/.github/workflows/format-tests.yml
+++ b/.github/workflows/format-tests.yml
@@ -31,23 +31,23 @@ jobs:
 
       - name: End to end test - Aerial Imagery
         run: |
-          docker run  -v "${{ runner.temp }}:/tmp/" topo-imagery python3 standardise_validate.py --from-file ./tests/data/aerial.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01 --gsd 10m --create-footprints=true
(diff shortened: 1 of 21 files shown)


blacha:
this is a breaking change, can it be tagged as breaking and explain the migration path is it just changing "1m" to "1" in all the CLIs?

Could this change be done in a backwards compatible way?

side note, all the e2e tests have highly uncommon GSDs 60m ? could a few of those be made more realistic, urban imagery is often in the 0.05m, 0.0125m or 0.075m sizes.

blacha (approved):

.github/workflows/format-tests.yml
```diff
@@ -31,23 +31,23 @@ jobs:
 
       - name: End to end test - Aerial Imagery
         run: |
-          docker run  -v "${{ runner.temp }}:/tmp/" topo-imagery python3 standardise_validate.py --from-file ./tests/data/aerial.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01 --gsd 10m --create-footprints=true
+          docker run  -v "${{ runner.temp }}:/tmp/" topo-imagery python3 standardise_validate.py --from-file ./tests/data/aerial.json --preset webp --target-epsg 2193 --source-epsg 2193 --target /tmp/ --collection-id 123 --start-datetime 2023-01-01 --end-datetime 2023-01-01 --gsd 10 --create-footprints=true
```
paulfouquet, line 34: Might be good to have one of the test keeping the `m`?

l0b0: I did: https://github.com/linz/topo-imagery/blob/df257c9b6c81224df5626097c4a906b353aed4aa/scripts/tests/collection_from_items_test.py#L55-L56

PR #1096, opened 2024-09-26
test: refactor fake collection metadata

paulfouquet:
### Motivation

The collection metadata used for testing is duplicated in the code base which makes it difficult to maintain.

### Modifications

Use the same collection metadata for all the related tests.

### Verification

ran pytest

### DIFF
--- a/scripts/stac/imagery/tests/conftest.py
+++ b/scripts/stac/imagery/tests/conftest.py
@@ -1,0 +1,22 @@
+from datetime import datetime
+from typing import Generator
+
+import pytest
+
+from scripts.stac.imagery.metadata_constants import CollectionMetadata
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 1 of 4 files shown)


scripts/stac/imagery/tests/conftest.py
```diff
@@ -0,0 +1,22 @@
+from datetime import datetime
+from typing import Generator
+
+import pytest
+
+from scripts.stac.imagery.metadata_constants import CollectionMetadata
+
+
+@pytest.fixture(autouse=True)
+def fake_collection_metadata() -> Generator[CollectionMetadata, None, None]:
```
l0b0, line 10: I believe the return type can be simplified to `Iterator[CollectionMetadata]` in general, when the second and third `Generator` types are `None`.

l0b0 (approved):
Nice! Included one simplification which is not necessary to do in this PR.

l0b0 (approved):

PR #1115, opened 2024-10-08
fix: providers should not be duplicated when having different roles TDE-1286

paulfouquet:
### Motivation

When a user add the same provider (same `name`) as a different role (for example, `--licensor='Maxar' --producer='Maxar'`) the provider should not be duplicated, but added with both role (for example: `{'name': 'Maxar', 'roles': ['producer', 'licensor']}`).

### Modifications

- merge provider list based on their name
- sort provider roles alphabetically (for readibility and conveniance) when added to the Collection

### Verification

automatic tests

### DIFF
--- a/scripts/stac/imagery/provider.py
+++ b/scripts/stac/imagery/provider.py
@@ -17,8 +17,7 @@ class Provider(TypedDict):
 
 
 def merge_provider_roles(providers: list[Provider]) -> list[Provider]:
-    """Merge the roles of the providers with the same name.
-    Does not manage duplicates in the final roles list.
+    """Remove duplicate providers based on name and merge their roles. See example below.
 
     Args:
         providers: a list of `Provider` objects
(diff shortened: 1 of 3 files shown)


scripts/stac/imagery/provider.py
```diff
@@ -14,3 +14,29 @@ class Provider(TypedDict):
     """Organization name"""
     roles: list[ProviderRole]
     """Organization roles"""
+
+
+def merge_provider_roles(providers: list[Provider]) -> list[Provider]:
+    """Merge the roles of the providers with the same name.
+    Does not manage duplicates in the final roles list.
```
amfage, line 21: I don't quite follow what this line means, can you explain to me like I am a 5 year old, please? :laughing:

paulfouquet: Yeah it's not great explaination. Could you please have another look?

amfage (dismissed):

amfage (approved):

PR #1147, opened 2024-10-25
refactor: use environment variable to get GDAL version

paulfouquet:
### Motivation

Simplify getting the gdal version.

### Modifications

Use system environment variable to store the gdal version value rather than calling the shell from Python.

### Verification

Ran E2E test

### DIFF
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -2,6 +2,9 @@
 
 set -o errexit
 
+GDAL_VERSION="$(gdalinfo --version)"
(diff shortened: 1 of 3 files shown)


amfage (approved):

schmidtnz (commented):
With these changes, a failed call to `gdalinfo` will ~~silently return an empty version string~~ fail early thanks to `set -o errexit`.
The python implementation logged and raised an exception of type `GDALExecutionException` if `gdalinfo` failed.

schmidtnz (approved):

PR #1152, opened 2024-10-30
feat!: add linz:slug to Collection TDE-1305

schmidtnz:
### Motivation

As a Basemaps developer I want the dataset name to be referenced within the collection.json so that I can look this up and use it as part of a link to the same dataset in Basemaps.

### Modifications

- Add `--linz-slug` argument to `collection_from_items.py`
- Add `linz:slug` attribute to `stac` property of `ImageryCollection`
- Update `ImageryCollection` class signature
- Update `create_collection` function signature
- Pass `linz_slug` from `collection_from_items.py` to `ImageryCollection`
- Tests updated

### Verification

Updated pytest tests to include `linz:slug` (and failure when the respective argument is missing).
Additional manual testing.

### DIFF
--- a/scripts/collection_from_items.py
+++ b/scripts/collection_from_items.py
@@ -109,7 +109,6 @@ def main(args: List[str] | None = None) -> None:
     arguments = parse_args(args)
     uri = arguments.uri
     collection_id = arguments.collection_id
-    linz_slug = arguments.linz_slug
 
     if not uri.startswith("s3://"):
         msg = f"uri is not a s3 path: {uri}"
@@ -178,7 +177,7 @@ def main(args: List[str] | None = None) -> None:
 
     collection = create_collection(
         collection_id=collection_id,
-        linz_slug=linz_slug,
+        linz_slug=arguments.linz_slug,
         collection_metadata=collection_metadata,
         producers=coalesce_multi_single(arguments.producer_list, arguments.producer),
         licensors=coalesce_multi_single(arguments.licensor_list, arguments.licensor),
--- a/scripts/collection_from_items.py
+++ b/scripts/collection_from_items.py
@@ -64,6 +64,13 @@ def parse_args(args: List[str] | None) -> Namespace:
         choices=["under development", "preview", "ongoing", "completed", "deprecated"],
     )
     parser.add_argument(
+        "--linz-slug",
+        dest="linz_slug",
+        help="linz:slug attribute for this dataset. E.g. bay-of-plenty_2018-2019_0.1m",
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
         "--producer",
         dest="producer",
         help="Imagery producer. Ignored if --producer-list passed with a semicolon delimited list.",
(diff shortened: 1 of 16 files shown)


schmidtnz:
BREAKING CHANGE: New required cli argument `--linz-slug` for `collection_from_items.py`. 
Not to be merged until respective changes in argo-tasks are ready.

scripts/collection_from_items.py
```diff
@@ -102,6 +109,7 @@ def main(args: List[str] | None = None) -> None:
     arguments = parse_args(args)
     uri = arguments.uri
     collection_id = arguments.collection_id
+    linz_slug = arguments.linz_slug
```
paulfouquet, line 112: This is called only once. You can call it L181 rather than assigning it to a variable.

scripts/stac/imagery/tests/conftest.py
```diff
@@ -21,3 +21,8 @@ def fake_collection_metadata() -> Iterator[CollectionMetadata]:
         "geographic_description": None,
     }
     yield collection_metadata
+
+
+@pytest.fixture(autouse=True)
```
paulfouquet, line 26: I don't think `autouse=True` [is necessary](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request)

paulfouquet (commented):
NITs

scripts/collection_from_items.py
l0b0: Basically, we generally inline single-use variables unless there's a good reason for the extra variable. One such good reason is if the original name and the preferred name in the current context have drifted, but we don't want to change the interface (yet).

schmidtnz: Agreed - was an oversight to leave that in.

scripts/stac/imagery/tests/conftest.py
schmidtnz: Interesting, it also appears to be safe to remove from L10.

schmidtnz:
Based on our discussion this morning, I've added a test for empty strings with the option to extend that further if needed.

scripts/collection_from_items.py
```diff
@@ -22,6 +22,12 @@ class NoItemsError(Exception):
     pass
 
 
+def linz_slug_type(value: str) -> str:
+    if not value:
```
l0b0, line 26: Do we also want to validate that it contains only [valid slug characters](https://github.com/linz/argo-tasks/blob/68940d582004f88350b30e8a6064dd1174e2631e/src/utils/slugify.ts#L15-L15)?

l0b0 (approved):

l0b0 (dismissed):

l0b0 (dismissed):

scripts/stac/imagery/tests/conftest.py
```diff
@@ -21,3 +21,8 @@ def fake_collection_metadata() -> Iterator[CollectionMetadata]:
         "geographic_description": None,
     }
     yield collection_metadata
+
+
+@pytest.fixture
+def fake_linz_slug() -> str:
```
paulfouquet, line 27: Can this be declared only once? 
It is duplicated in the code base.

l0b0: You could also convert it to not be a fixture, then it can be imported from anywhere.

schmidtnz: moved `conftest.py` to parent folder

paulfouquet (approved):

PR #1160, opened 2024-11-04
feat: keep Collection properties.created date when resupplying

l0b0:
### Motivation

Ensure that the STAC Collection `properties.created` datetime is not changed when resupplying.

### Modifications

- Add `--published-path` parameter to `collection_from_items.py`, to get created datetime on resupply.
- Add `--current-datetime` parameter to `collection_from_items.py` to use that instead of the actual clock time for created/updated datetimes.

### DIFF
--- a/scripts/collection_from_items.py
+++ b/scripts/collection_from_items.py
@@ -32,6 +32,12 @@ def parse_args(args: List[str] | None) -> Namespace:
     parser.add_argument("--uri", dest="uri", help="s3 path to items and collection.json write location", required=True)
     parser.add_argument("--collection-id", dest="collection_id", help="Collection ID", required=True)
     parser.add_argument(
+        "--odr-url",
+        dest="odr_url",
+        help="The path of the published dataset. Example: 's3://nz-imagery/wellington/porirua_2024_0.1m/rgb/2193/'",
+        required=False,
+    )
+    parser.add_argument(
         "--category",
         dest="category",
         help="Dataset category description",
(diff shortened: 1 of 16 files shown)


scripts/collection_from_items.py
```diff
@@ -31,6 +31,12 @@ def parse_args(args: List[str] | None) -> Namespace:
     parser = argparse.ArgumentParser()
     parser.add_argument("--uri", dest="uri", help="s3 path to items and collection.json write location", required=True)
     parser.add_argument("--collection-id", dest="collection_id", help="Collection ID", required=True)
+    parser.add_argument(
+        "--published-path",
+        dest="published_path",
+        help="The path of the published dataset. Example: 's3://nz-imagery/wellington/porirua_2024_0.1m/rgb/2193/'",
```
schmidtnz, line 37: In other places, this was called `odr-url`. Could be good to keep it consistent.

scripts/stac/imagery/create_stac.py
```diff
@@ -82,6 +87,15 @@ def create_collection(
     return collection
 
 
+def get_providers(licensors: list[str], producers: list[str]) -> list[Provider]:
+    providers: list[Provider] = []
+    for producer_name in producers:
+        providers.append({"name": producer_name, "roles": [ProviderRole.PRODUCER]})
+    for licensor_name in licensors:
+        providers.append({"name": licensor_name, "roles": [ProviderRole.LICENSOR]})
+    return providers
```
schmidtnz, line 96: Is this the desired output for `get_providers(["Prov1", "Prov2"], ["Prov2"])`?
`[{'name': 'Prov2', 'roles': ['producer']}, {'name': 'Prov1', 'roles': ['licensor']}, {'name': 'Prov2', 'roles': ['licensor']}]`
Or would we be expecting something like this instead:
`[{'name': 'Prov2', 'roles': ['producer', 'licensor']}, {'name': 'Prov1', 'roles': ['licensor']}]`
?

l0b0: Good point, but I think the current code is identical to the original. I just had to reduce the number of local variables for pylint.

schmidtnz (approved):

PR #1161, opened 2024-11-05
test: Create fixtures for STAC metadata

l0b0:


l0b0:
This was just a showcase.

PR #1180, opened 2024-11-20
feat: set update-dates on STAC item TDE-1298

schmidtnz:
### Motivation

As a public bucket Imagery or Elevation consumer I want to know what new datasets have been added, what datasets have had changes, what metadata has changed, so that I can keep my data and processes in sync with data referenced from the public buckets.
When an ODR path is provided to the imagery-standardising workflow, look up previously published STAC Item documents (by filename) and use them if the files are unchanged (file size and checksum).

### Modifications

- **Entry point `standardise_validate.py`**: 
  - Added `--current-datetime` argument (defaults to "today" to avoid making a breaking change).
  - Added `--odr-url` optional argument.
- **ImageryItem**:
  - Added @classmethod to load an existing _Item_ from file/s3.
  - Added `set_checksum()` method which will set the `asset.visual.file:checksum` and associated attributes if the given checksum differs from the current checksum.
  - Removed `now_string` from class init. Created/updated dates now set through respective `asset` properties.
  - Associated changes to reflect updated signature in all tests. 
  - ....

### Verification

Added and modified tests for `item`, `collection`, `create_item`, `create_stac` and `collection_from_items`.


scripts/stac/imagery/item.py
```diff
@@ -26,17 +28,59 @@
...
+        Args:
+            file_name: The s3 URL or local path of the file to load.
+
+        Returns:
+            ImageryItem: The new ImageryItem.
+        """
+        file_content = read(file_name)
+        stac_dict_from_s3 = json.loads(file_content.decode("UTF-8"))
+        if "bbox" in stac_dict_from_s3:  # tuple() does not serialise well into json
+            stac_dict_from_s3["bbox"] = tuple(stac_dict_from_s3["bbox"])
```
l0b0, line 55: Nit: We can use the walrus operator to avoid a lookup, like this:

```
        if (bbox := stac_dict_from_s3.get("bbox")) is not None:
            stac_dict_from_s3["bbox"] = tuple(bbox)
```

It *is* more complex syntax, though, so feel free to ignore.

scripts/stac/imagery/item.py
```diff
@@ -26,17 +28,59 @@
...
+        new_item = cls(
+            id_=stac_dict_from_s3["id"],
+            stac_asset=stac_dict_from_s3["assets"]["visual"],
+            stac_processing=stac_dict_from_s3["properties"],
+        )
+        new_item.stac = stac_dict_from_s3
+
+        return new_item
+
+    def set_checksum(self, file_content_checksum: str, stac_processing_data: STACProcessing) -> None:
```
l0b0, line 65: Rather than passing a `STACProcessing` object, I'd probably just pass the attributes of the object which are needed for this method. That way this method doesn't need to know about the structure of `STACProcessing`, and we refactor it easier.

paulfouquet, line 65: This function is called `set_checksum()` but does other important things than just setting the `file:checksum`. Have you thought about renaming it or pulling out some of the logic that is not related to setting the checksum?

schmidtnz: I feel like all of it is _related_ to setting the checksum (similar to how when a file gets written, its "modified" date also updates without the need to call "set modified date" separately). 

Do you have a suggestion for a better name? `update_checksum_if_different_with_metadata` ? Feels long.

l0b0: You could do something like `for key, value in stac_processing_data.items() if key != "processing:datetime":`, so that this method doesn't need to know anything else about the structure of the object?

l0b0: How about `update_checksum_related_metadata`?

l0b0 (dismissed):

PR #1220, opened 2024-12-17
fix: handle existing derived_from links when Item is updated TDE-1356

paulfouquet:
### Motivation

When a published Item with `derived_from` links is resupplied, its existing list of `derived_from` links may have changed with the resupply. The current implementation would duplicate existing `derived_from` links and left obsolete ones. It will also remove any `derived_from` links in case of the Item is not derived from other Items anymore.

### Modifications

Remove existing `derived_from` links to the STAC Item before processing the current `derived_from` links list.

### Verification

Pytest
Argo Workflows

### DIFF
--- a/scripts/stac/imagery/create_stac.py
+++ b/scripts/stac/imagery/create_stac.py
@@ -130,6 +130,9 @@ def create_item(
         gdalinfo_result = gdal_info(asset_path)
 
     if derived_from is not None:
+        # Remove existing derived_from links in case of resupply
+        item.stac["links"] = [link for link in item.stac["links"] if link["rel"] != "derived_from"]
+
         for derived in derived_from:
             derived_item_content = read(derived)
             derived_stac = json.loads(derived_item_content.decode("UTF-8"))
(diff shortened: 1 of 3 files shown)


scripts/stac/imagery/create_stac.py
```diff
@@ -130,6 +130,9 @@ def create_item(
         gdalinfo_result = gdal_info(asset_path)
 
     if derived_from is not None:
+        # Remove existing derived_from links in case of resupply
+        item.stac["links"] = [link for link in item.stac["links"] if link["rel"] != "derived_from"]
```
schmidtnz, line 134: might be better to move this outside of `if derived_from is not None:`.

schmidtnz (changes requested):

scripts/stac/imagery/create_stac.py
paulfouquet: As discussed, from business pov there should not be an issue as I don't think an existing Item that is derived from would become an Item from its own, but technically that can happen.

schmidtnz (approved):

PR #1224, opened 2024-12-19
feat: update existing Collection TDE-1359

paulfouquet:
### Motivation

When a provider supplies an update for an already published dataset, we would like to update the Collection accordingly to the new files supplied.

> **_NOTE:_**  The current workaround is to merge the existing data with the new supply in order to process them an recreate a new Collection. This not efficient.

### Modifications

- Load a STAC Collection from its published location
- Update existing Item links and add new Item links
- Recreate spatial and temporal extents based on new Item list
- Update existing capture-area

### Verification

- Unit tests
- Manual tests
- Argo Workflows tests


scripts/stac/imagery/tests/create_stac_test.py
```diff
@@ -479,3 +493,84 @@ def test_create_item_when_resupplying_with_changed_asset_file(subtests: SubTests
...
+                "properties": {"start_datetime": "2024-09-02T12:00:00Z", "end_datetime": "2024-09-02T12:00:00Z"},
+            },
+            {"type": "Feature", "id": "item_a"},
+            {"type": "Feature", "id": "item_c"},
+        ]
+
+    with subtests.test("links"):
+        assert collection.stac["links"] == [{"rel": "self", "href": "./collection.json"}]
+
+    with subtests.test("extent"):
```
paulfouquet, line 575: Note to myself, needs to actuall add an extent to the collection before running `prepare_resupply`

scripts/stac/imagery/collection.py
```diff
@@ -105,7 +143,9 @@ def add_capture_area(self, polygons: list[BaseGeometry], target: str, artifact_t
             artifact_target: location where the capture-area.geojson artifact file will be saved.
             This is useful for Argo Workflow in order to expose the file to the user for testing/validation purpose.
         """
-
+        # If published dataset update, merge the existing capture area with the new one
```
amfage, line 146: Do we need to check for a case where the original dataset does not have a capture area but the newly supplied files are processed with one, and not create a capture area in that case?

paulfouquet: I agree we need to check that

amfage (approved):

PR #1253, opened 2025-01-21
feat: generate hillshade TDE-1369

paulfouquet:
### Motivation

LINZ wants to offer some standard "out-of-the-box" hillshades to the elevation data consumers. This feature gives the ability to generate two standard hillshades from existing DEM TIFF(s):
<div align="center">

| [:world_map: Monodirectional or "Greyscale" hillshade](https://basemaps.linz.govt.nz/@-40.6547814,176.2198739,z3.69?style=23-test-hillshade-7w5ph&i=23-test-hillshade-7w5ph&tileMatrix=NZTM2000Quad&config=6sUk4Za2WKysFTtUuDLY8VZKUUvfvkXoNXGCqB3PC6uKQH1kbhniZaS5JVmTxb8WBZ1RBvwbK3xk8XrDdC37MwSj1sn3XBLwcFxEJf1mR6fHJdoP8LYt624nphascfFHdL8v83CksZqqWKvhcGBLTpHJsL1PeuHMMuCGqUG6eNUAu9FWxajyLD&debug=true&debug.layer=basemaps-23-test-hillshade-7w5ph-color-ramp)  | [:world_map: "Igor" hillshade](https://basemaps.linz.govt.nz/@-40.1252942,174.5734854,z5?style=24-test-hillshade-fpps6&i=24-test-hillshade-fpps6&tileMatrix=NZTM2000Quad&config=6sUk4Za2WKysFTtUuDLY8VZKUUvfvkXoNXGCqB3PC6uKQH1kbhniuZFFm2BPvcgeALCZJgJ3NRK6Y3kgEH7pKGK5LadeRw3LZDiivVeEE4zGRufribx7q1EAHzMgwvF6vQ6ZqePJHk4Gb54aXzBs4Ey8WXhgp2hUARXq1F6o1AzJ9fHe3anphw&debug=true&debug.layer=basemaps-24-test-hillshade-fpps6-color-ramp) |
| ------------- | ------------- |
|   |  |

</div>

### Modifications

- add a script `generate_hillshade.py`
```
usage: generate_hillshade.py [-h] --from-file FROM_FILE --preset
                             {greyscale,igor} --target TARGET [--force]

options:
  -h, --help            show this help message and exit
  --from-file FROM_FILE
                        Specify the path to a json file containing the input
                        tiffs. Format: [{'output': 'tile1', 'inputs':
                        ['path/input1.tiff', 'path/input2.tiff']}]
  --preset {greyscale,igor}
                        Type of hillshade to generate.
  --target TARGET       Specify the path to save the generated hillshade to.
  --force               Regenerate the hillshade TIFF files if already exist.
                        Defaults to False.
```

### Verification

Manual, automated end-to-end tests, and Argo Workflows runs.


scripts/generate_hillshade.py
```diff
@@ -0,0 +1,203 @@
...
+    return results
+
+
+def main() -> None:
+    start_time = time_in_ms()
+    parser = get_args_parser()
+    arguments = parser.parse_args()
+
+    if not arguments.no_stac and not arguments.collection_id:
+        parser.error(
```
schmidtnz, line 159: What are your thoughts about this - Having the no_stac arg here seems redundant. Would it be sufficient to only create STAC if collection_id is set?

paulfouquet: Good idea.
After our discussion about the overall process that would call that script, I'm now thinking to remove (again) the STAC creation here...

scripts/generate_hillshade.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Returns:
+        The filename of the hillshade TIFF file if created.
+    """
+    hillshade_file_name = tile.output + ".tiff"
+
+    hillshade_file_path = os.path.join(target_output, hillshade_file_name)
+
+    # Already processed can skip processing
+    if exists(hillshade_file_path):
```
schmidtnz, line 69: Under what circumstances would this actually trigger? Files are being processed into a different scratch directory (or even a container local temp dir) on each workflow run, right?

scripts/generate_hillshade.py
```diff
@@ -0,0 +1,158 @@
...
+        preset: a `HillshadePreset` to use. See `gdal.gdal_presets.py`.
+        target_output: path where the output files need to be saved to. Defaults to "/tmp/".
+        force: overwrite existing output file. Defaults to False.
+
+    Returns:
+        The filename of the hillshade TIFF file if created.
+    """
+    hillshade_file_name = tile.output + ".tiff"
+
+    hillshade_file_path = os.path.join(target_output, hillshade_file_name)
```
schmidtnz, line 66: Do you think it might be better to include the `preset`, too? I'm undecided.

scripts/generate_hillshade.py
paulfouquet: If a workflow task using this script fails for an intermittent issue (example: network related, pod killed etc.), it allows to **"resume"** (via retry) the failed task without restarting it or the entire workflow.

scripts/generate_hillshade.py
paulfouquet: I feel the specific naming of a file should not be handled by this script, and should just follow the regular convention of [TILENAME].tiff

scripts/generate_hillshade.py
paulfouquet: > or even a container local temp dir
At the end of the process of this script, they are saved to the `target`, which is usually a s3 location.

scripts/generate_hillshade.py
schmidtnz: sorry, not suggesting to change the file name necessarily. 

`target_output` defaults to `/tmp/` and is likely going to be a unique path containing some random string. 
The workflow will output two hillshades for each tile, and with them going to "random" output folders, will it be easy enough to differentiate between each of them (when needed, e.g. debugging or manual QC) without any distinction in the name/path?

The tile may still be called `xx00.tiff`,  but it could be in a `target_output/preset/` sub-folder?

paulfouquet: If we are thinking Argo Workflows, the target would be defined by the caller of the task that run this script. Most likely something like `s3://linz-workflows-scratch/[date]/[workflow-id]/flat/`

scripts/gdal/gdal_commands.py
```diff
@@ -195,3 +196,34 @@ def get_ascii_translate_command() -> list[str]:
...
+        a `gdaldem` command
+    """
+    gdal_command: list[str] = [
+        "gdaldem",
+        "hillshade",
+        "-compute_edges",
+        "-of",
+        "COG",
+        "-co",
+        "compress=lerc",
```
schmidtnz, line 217: command as per jira ticket also explicitly sets `overview_compress=lerc` and `max_z_error_overview=0` (equivalent to defaults)

schmidtnz (approved):

PR #1268, opened 2025-01-30
feat: add '-resolution highest' argument to gdalbuildvrt

schmidtnz:


paulfouquet:
Did you want to modify the standardise_validate.py script to allow user to pass this new parameter?

schmidtnz:
> Did you want to modify the standardise_validate.py script to allow user to pass this new parameter?

sorry, makes sense, for some reason I thought that would come from generate_hillshade

PR #1269, opened 2025-01-31
feat: add resolution argument to gdalbuildvrt TDE-1383

schmidtnz:
### Motivation

To generate a national combined hillshade product, we need to be able to combine inputs of differing resolutions and ensure consistent output.
The default settings of `gdalbuildvrt` will average all input resolutions. This change implements a more suitable setting for our use-case.

### Modifications
Add optional `resolution` argument to  `get_build_vrt_command()`, which will in turn add `-resolution user -tr <xres> <yres>` arguments to the respective GDAL command.

`standardising()` now accepts a list `scale_to_resolution = [x_res, y_res]` to support the above.

### Verification

Manual test run with combined DEM hillshades.

### DIFF
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -26,9 +26,9 @@ def str_to_bool(value: str) -> bool:
 
 
 def str_to_list_or_none(values: str) -> list[Decimal] | None:
-    if values is None or values == "":
+    if not values:
         return None
-    result = [Decimal(val) for val in values.split(",")]
+    result = [Decimal(value) for value in values.split(",")]
     if len(result) != 2:
         raise argparse.ArgumentTypeError(f"Invalid list - must be blank or exactly 2 values x,y. Received: {values}")
     return result
@@ -160,13 +160,7 @@ def main() -> None:
 
     gdal_version = os.environ["GDAL_VERSION"]
 
-    tiff_files = run_standardising(
-        tile_files,
-        standardising_config,
-        concurrency,
-        gdal_version,
-        arguments.target,
-    )
+    tiff_files = run_standardising(tile_files, standardising_config, concurrency, gdal_version, arguments.target)
 
     if len(tiff_files) == 0:
         get_log().info("no_tiff_to_process", action="standardise_validate", reason="skipped")
--- a/scripts/standardise_validate.py
+++ b/scripts/standardise_validate.py
@@ -14,7 +14,7 @@ from scripts.files.fs import exists, write
 from scripts.gdal.gdal_helper import get_srs, get_vfs_path
 from scripts.json_codec import dict_to_json_bytes
 from scripts.stac.imagery.create_stac import create_item
-from scripts.standardising import run_standardising
+from scripts.standardising import StandardisingConfig, run_standardising
 
 
 def str_to_bool(value: str) -> bool:
@@ -30,7 +30,7 @@ def str_to_list_or_none(values: str) -> list[Decimal] | None:
         return None
     result = [Decimal(val) for val in values.split(",")]
     if len(result) != 2:
-        raise argparse.ArgumentTypeError(f"Invalid list (must be blank or exactly 2 values): {values}")
+        raise argparse.ArgumentTypeError(f"Invalid list - must be blank or exactly 2 values x,y. Received: {values}")
     return result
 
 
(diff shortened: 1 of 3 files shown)


paulfouquet:
> Note: This may lead to **different outputs than previously**, and will lead to fewer errors when processing data (basemaps checks fail if output tiles do not all have the same resolution).

Some finding while testing?

schmidtnz:
> > Note: This may lead to **different outputs than previously**, and will lead to fewer errors when processing data (basemaps checks fail if output tiles do not all have the same resolution).
> 
> Some findings while testing?

as discussed, this was not likely to have been an issue with the current script logic and human workflows, however it went against one of the future features (validating if GSD matches TIFF files), so this was moved to a separate CLI argument.

scripts/standardise_validate.py
```diff
@@ -24,6 +25,15 @@ def str_to_bool(value: str) -> bool:
     raise argparse.ArgumentTypeError(f"Invalid boolean (must be exactly 'true' or 'false'): {value}")
 
 
+def str_to_list_or_none(values: str) -> list[Decimal] | None:
+    if values is None or values == "":
```
paulfouquet, line 29: If you don't expect a `0`, you could simplify?

```suggestion
    if not values:
```

scripts/standardising.py
```diff
@@ -26,54 +29,73 @@
...
+    gsd: Decimal
+    create_footprints: bool
+    cutline: str | None
+    scale_to_resolution: list[Decimal] | None = None
+
+    def __post_init__(self) -> None:
+        if self.cutline is not None and not self.cutline.endswith((".fgb", ".geojson")):
+            raise ValueError(f"Only .fgb or .geojson cutlines are supported: {self.cutline}")
+        if self.scale_to_resolution is not None and len(self.scale_to_resolution) != 2:
+            raise ValueError(f"Scale_to_resolution must be exactly two items [xres, yres]: {self.scale_to_resolution}")
```
paulfouquet, line 55:
```suggestion
            raise ValueError(f"scale_to_resolution must be exactly two items [xres, yres]: {self.scale_to_resolution}")
```

scripts/standardising.py
```diff
@@ -26,54 +29,73 @@
...
+
+
 def run_standardising(
-    todo: list[TileFiles],
-    preset: str,
-    cutline: str | None,
+    files_to_process: list[TileFiles],
```
paulfouquet, line 59: Tiles might be even more correct?

```suggestion
    tiles_to_process: list[TileFiles],
```

scripts/standardising.py
```diff
@@ -85,164 +107,220 @@ def run_standardising(
...
+            footprint_file_path = os.path.join(target_output, f"{files.output}{SUFFIX_FOOTPRINT}")
+            write(footprint_file_path, read(temp_footprint), content_type=ContentType.GEOJSON.value)
+
+        # Copy the final version of the working / temp file to the desired destination
+        write(standardised_file_path, read(current_working_file), content_type=ContentType.GEOTIFF.value)
+
+    return tiff
+
+
+def generate_prj_tfw_sidecars(tiff: FileTiff, target_path: str) -> list[str]:
```
paulfouquet, line 215: I don't think that is a "generation", it must be a "copy" or "get". The system tries to download any sidecar files (files that are associated to / have been delivered with the TIFFs), assuming they have the same basename than the TIFF with extensions `.pjr` and `.tfw`.

scripts/standardising.py
```diff
@@ -85,164 +107,220 @@ def run_standardising(
...
+    get_log().info("Running GDAL", command=command, input_file=input_file, output_file=target_file)
+
+    # Need GDAL to write to temporary location so no broken files end up in the done folder.
+    run_gdal(command, input_file=input_file, output_file=target_file)
+
+    return target_file
+
+
+def check_tiff_empty(current_working_file: str) -> bool:
+    """Validate if the TIFF output is empty."""
```
paulfouquet, line 303:
```suggestion
    """Validate if the TIFF output is not empty."""
```

paulfouquet (commented):
It's nice to finally see the standardising function getting refactored!
A few minor comments.

scripts/standardising.py
schmidtnz: You're absolutely right. I keep misreading what the write* functions do.

scripts/standardise_validate.py
schmidtnz: This is user input, ... I don't _expect_ `0` but eventually I will receive `0` 🤣

schmidtnz: With this change, the impact will be that if someone specifies 0 as `scale_to_resolution` the script will not raise an exception - which is incorrect, but probably acceptable.

scripts/standardising.py
schmidtnz: This will return true if empty, false if not empty, so I'd like to keep the doc string positive.
With "not empty" in the doc string I would expect it to return true (success) if not empty, and false (fail) if empty.
I've reworded the doc string to match the function name (Check whether ...)

paulfouquet (approved):

PR #1286, opened 2025-02-13
fix(standardise-validate): allow empty string from Argo for start and end dates TDE-1346

amfage:
### Motivation

Argo Workflows supplies single quotes for optional parameters, which are not valid dates. These are handled later in the `standardise_validate.py` code.

### Modifications

Allow an empty string to be accepted by the command line parser for the start and end dates.

### Verification

Added unit tests. Tested from Argo Workflows using a PR container.

### DIFF
--- a/scripts/cli/tests/cli_helper_test.py
+++ b/scripts/cli/tests/cli_helper_test.py
@@ -1,6 +1,9 @@
+from datetime import datetime
+
+from pytest import raises
 from pytest_subtests import SubTests
 
-from scripts.cli.cli_helper import TileFiles, coalesce_multi_single, get_tile_files, parse_list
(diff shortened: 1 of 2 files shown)


scripts/cli/tests/cli_helper_test.py
```diff
@@ -71,3 +74,17 @@ def test_coalesce_nothing() -> None:
     single_item = ""
     coalesced_list = coalesce_multi_single(multi_items, single_item)
     assert coalesced_list == []
+
+
+def test_valid_date_empty_string() -> None:
```
paulfouquet, line 79: You can also create [doctests](https://docs.python.org/3/library/doctest.html) for this kind of unit tests.

paulfouquet (approved):
A NIT and subjective comment

PR #1287, opened 2025-02-17
fix: handle optional flags from Argo TDE-1346

amfage:
### Motivation
Argo Workflows supplies single quotes where parameters are not supplied, which can happen if a workflow template is used by multiple workflows. This breaks optional parameters in argparse.

### Modifications

Convert Argo string to boolean to control adding the capture dates file and the title suffix, in line with how the capture area creation (footprint) is handled.

### Verification

Ran workflow in Argo and locally on the command line.

### DIFF
--- a/scripts/cli/cli_helper.py
+++ b/scripts/cli/cli_helper.py
@@ -136,3 +136,20 @@ def str_to_gsd(value: str) -> Decimal:
         )
 
     return Decimal(number_value)
+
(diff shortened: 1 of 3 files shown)


paulfouquet (commented):
Could you add some tests for these utility functions part of this PR?

paulfouquet (approved):

PR #1288, opened 2025-02-20
feat: normalize capture-area geometry TDE-1359

paulfouquet:
### Motivation

When an existing capture-area geometry is being modified (polygon being removed/added), as a result the polygons points can be moved around. When the overall geometry ends up being the same (no visual change), we prefer the capture-area geometry to not show a change (difference between existing and new/re-processed one).

### Modifications

- Use [shapely.normalize()](https://shapely.readthedocs.io/en/2.0.3/reference/shapely.normalize.html) to keep the geometry consistent.

What does `shapely.normalize()`?

According to [the official documentation](https://shapely.readthedocs.io/en/latest/geometry.html#canonical-form):

> - the coordinates of exterior rings follow a clockwise orientation and interior rings have a counter-clockwise orientation
> - the starting point of rings is lower left
> - elements in collections are ordered by geometry type: by descending dimension and multi-types first (MultiPolygon, Polygon, MultiLineString, LineString, MultiPoint, Point). Multiple elements from the same type are ordered from right to left and from top to bottom.

Because, "_the coordinates of exterior rings follow a clockwise orientation  and interior rings have a counter-clockwise orientation_" is going against [RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6) - "_A linear ring MUST follow the right-hand rule with respect to the area it bounds, i.e., exterior rings are counterclockwise, and holes are clockwise._", we need to apply `orient()` after `normalize()` the geometry.

### Verification

- Unit tests
- Other tests to come  in https://github.com/linz/topo-imagery/pull/1224

### DIFF
--- a/scripts/stac/imagery/tests/capture_area_test.py
+++ b/scripts/stac/imagery/tests/capture_area_test.py
@@ -2,8 +2,8 @@ from decimal import Decimal
 from sys import float_info
 from typing import cast
 
-from shapely import MultiPolygon, get_exterior_ring, is_ccw
-from shapely.geometry import Polygon, shape
+from shapely import get_exterior_ring, is_ccw
+from shapely.geometry import MultiPolygon, Polygon, shape
 
 from scripts.stac.imagery.capture_area import generate_capture_area, merge_polygons, to_feature
 
--- a/scripts/stac/imagery/tests/capture_area_test.py
+++ b/scripts/stac/imagery/tests/capture_area_test.py
@@ -5,11 +5,7 @@ from typing import cast
 from shapely import MultiPolygon, get_exterior_ring, is_ccw
 from shapely.geometry import Polygon, shape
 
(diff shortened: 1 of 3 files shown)


scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -2,8 +2,8 @@
 from sys import float_info
 from typing import cast
 
-from shapely import get_exterior_ring, is_ccw
-from shapely.geometry import MultiPolygon, Polygon, shape
+from shapely import MultiPolygon, get_exterior_ring, is_ccw
+from shapely.geometry import Polygon, shape
```
amfage, line 6:
```suggestion
from shapely import MultiPolygon, Polygon, get_exterior_ring, is_ccw
from shapely.geometry import shape
```

amfage (approved):

PR #1305, opened 2025-03-21
refactor: make CollectionMetadata a dataclass TDE-1443

paulfouquet:
### Motivation

Using a `dataclass` gives advantages such as type safety at runtime, attributes access vs dictionary lookups, supports default values, IDE hints are better, etc.

### Modifications

- Convert `CollectionMetadata` from a `TypedDict` to a `dataclass`
The caveat is that the linter complains about too many instance attributes.
I've thought about splitting it up into few other dataclass, but it makes it more readable and maintainable.

### Verification

Automatic tests

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -71,9 +71,9 @@ class ImageryCollection:
             "license": "CC-BY-4.0",
             "links": [{"rel": "self", "href": "./collection.json", "type": "application/json"}],
             "providers": [],
-            "linz:lifecycle": metadata["lifecycle"],
-            "linz:geospatial_category": metadata["category"],
-            "linz:region": metadata["region"],
+            "linz:lifecycle": metadata.lifecycle,
+            "linz:geospatial_category": metadata.category,
+            "linz:region": metadata.region,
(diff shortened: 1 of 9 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -71,19 +71,19 @@ def __init__(
...
 
         # Optional metadata
-        if event_name := metadata.get("event_name"):
+        if event_name := metadata.event_name:
             self.stac["linz:event_name"] = event_name
-        if geographic_description := metadata.get("geographic_description"):
+        if geographic_description := metadata.geographic_description:
             self.stac["linz:geographic_description"] = geographic_description
 
         # If the providers passed has already a LINZ provider: add its default roles to it
```
schmidtnz, line 89: This might be redundant when using `merge_provider_roles()` ? Can we just add LINZ and it will get deduplicated?

schmidtnz (approved):

scripts/stac/imagery/collection.py
paulfouquet: We can refactor that in a separate PR. I think it's out of scope for that one.

PR #1309, opened 2025-03-26
refactor: `ImageryCollection.from_file()` only handles loading the STAC metadata TDE-1443

paulfouquet:
### Motivation

`ImageryCollection.from_file()` should only handle loading the STAC metadata from the existing Collection file and not doing any other logic.

### Modifications

- Move extra logic where loading the collection is called from.

### Verification

Automated tests

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -45,7 +45,6 @@ WARN_NO_PUBLISHED_CAPTURE_AREA = "no_published_capture_area"
 
 class ImageryCollection:
     stac: dict[str, Any]
-    metadata: CollectionMetadata
     capture_area: dict[str, Any] | None = None
     published_location: str | None = None
 
@@ -104,12 +103,11 @@ class ImageryCollection:
         self.add_providers(merge_provider_roles(providers))
 
     @classmethod
-    def from_file(cls, path: str, metadata: CollectionMetadata) -> "ImageryCollection":
+    def from_file(cls, path: str) -> "ImageryCollection":
         """Load an ImageryCollection object from a STAC Collection file.
 
         Args:
             path: The path to the STAC Collection file to load.
-            metadata: The metadata of the Collection.
 
         Returns:
             The ImageryCollection loaded from the file.
@@ -117,16 +115,7 @@ class ImageryCollection:
         file_content = read(path)
         stac_from_file = json.loads(file_content.decode("UTF-8"))
         collection = cls.__new__(cls)
-        if metadata.collection_id != stac_from_file["id"]:
-            raise ValueError(f"Collection ID mismatch: input={metadata.collection_id} != existing={stac_from_file['id']}")
-        if metadata.linz_slug != stac_from_file["linz:slug"]:
-            get_log().warn(
-                f"Collection LINZ slug mismatch: input={metadata.linz_slug} != "
-                f"existing={stac_from_file['linz:slug']}."
-                "Keeping existing Collection linz_slug."
-            )
         collection.stac = stac_from_file
-        collection.metadata = metadata
         collection.published_location = os.path.dirname(path)
         capture_area_path = os.path.join(collection.published_location, CAPTURE_AREA_FILE_NAME)
         # Some published datasets may not have a capture-area.geojson file (TDE-988)
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -116,6 +116,11 @@ class ImageryCollection:
         stac_from_file = json.loads(file_content.decode("UTF-8"))
         collection = cls.__new__(cls)
         collection.stac = stac_from_file
+        collection.published_location = os.path.dirname(path)
+        capture_area_path = os.path.join(collection.published_location, CAPTURE_AREA_FILE_NAME)
+        # Some published datasets may not have a capture-area.geojson file (TDE-988)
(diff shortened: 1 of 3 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -103,40 +103,19 @@ def __init__(
...
-                "Existing Collection linz_slug will be used."
-            )
-            metadata.linz_slug = stac_from_file["linz:slug"]
-        collection = cls(
-            metadata=metadata,
-            created_datetime=stac_from_file["created"],
-            updated_datetime=stac_from_file["updated"],
-        )
-        # Override STAC from the original collection
+        collection = cls.__new__(cls)
```
paulfouquet, line 117: Could set the `published_location` here. Might have some typing issues where it is called as `published_location` can be `None`?

scripts/stac/imagery/create_stac.py
```diff
@@ -65,9 +65,25 @@ def create_collection(  # pylint: disable=too-many-arguments
...
+                f"Collection LINZ slug mismatch: input={collection_metadata.linz_slug} != "
+                f"existing={collection.stac['linz:slug']}."
+                "Keeping existing Collection linz_slug."
+            )
+        collection.stac["updated"] = current_datetime
+        collection.published_location = os.path.dirname(existing_collection_path)
+        capture_area_path = os.path.join(collection.published_location, CAPTURE_AREA_FILE_NAME)
+        # Some published datasets may not have a capture-area.geojson file (TDE-988)
+        if exists(capture_area_path):
+            collection.capture_area = json.loads(read(capture_area_path))
```
schmidtnz, line 86: Would this be seen as part of producing a correct representation of the file content as an `ImageryCollection` object? Might be better to keep this back inside the from_file function?

scripts/stac/imagery/create_stac.py
```diff
@@ -65,9 +65,25 @@ def create_collection(  # pylint: disable=too-many-arguments
...
+                f"Collection ID mismatch: input={collection_metadata.collection_id} != existing={collection.stac['id']}"
+            )
+        if collection_metadata.linz_slug != collection.stac["linz:slug"]:
+            get_log().warn(
+                f"Collection LINZ slug mismatch: input={collection_metadata.linz_slug} != "
+                f"existing={collection.stac['linz:slug']}."
+                "Keeping existing Collection linz_slug."
+            )
+        collection.stac["updated"] = current_datetime
+        collection.published_location = os.path.dirname(existing_collection_path)
```
schmidtnz, line 82: Would this be seen as part of producing a correct representation of the file content as an `ImageryCollection` object? Might be better to keep this back inside the from_file function?

scripts/stac/imagery/collection.py
```diff
@@ -103,40 +103,19 @@ def __init__(
...
         collection.stac = stac_from_file
```
schmidtnz, line 118: Do we need to set any other properties?

scripts/stac/imagery/collection.py
schmidtnz: sorry missed your comment. I think it would be better to set `published_location` (and other properties) here to build a "valid" `ImageryCollection` before returning it, but stick to the values coming from file.

scripts/stac/imagery/collection.py
paulfouquet: That is currently the only property that is set by the `__init__()` constructor.
However, it may be a good idea in the future to have a STAC update option, in case there is a change in the STAC document structure that would need to happen in the loaded Collection. I feel `pystac` has [something similar](https://github.com/stac-utils/pystac/blob/main/pystac/serialization/migrate.py).

scripts/stac/imagery/collection.py
```diff
@@ -103,40 +103,24 @@ def __init__(
...
-        if metadata.collection_id != stac_from_file["id"]:
-            raise ValueError(f"Collection ID mismatch: {metadata.collection_id} != {stac_from_file['id']}")
-        if metadata.linz_slug != stac_from_file["linz:slug"]:
-            get_log().warn(
-                f"Collection LINZ slug mismatch: {metadata.linz_slug} != {stac_from_file['linz:slug']}."
-                "Existing Collection linz_slug will be used."
-            )
-            metadata.linz_slug = stac_from_file["linz:slug"]
-        collection = cls(
-            metadata=metadata,
```
schmidtnz, line 127: we need to consider that the init also sets
`        self.metadata = metadata`
Do we need to assemble a metadata object when loading from file?
And on a side note, the `CollectionMetadata` class should only really contain properties that are actual metadata fields in the STAC document. 
Something like the optional `geographic_description` seems valid, but `historic_survey_number` only ends up in the title, not in a STAC property. --> might be better left to a future refactor, just to keep in mind here that we can't get the full `metadata` object from what we have in the STAC.

paulfouquet: What you stated here:

> Do we need to assemble a metadata object when loading from file?

is a fix that will happen in a future PR that is to resolve TDE-1444.

As discussed, I agree with the following:

> And on a side note, the CollectionMetadata class should only really contain properties that are actual metadata fields in the STAC document.
Something like the optional geographic_description seems valid, but historic_survey_number only ends up in the title, not in a STAC property. --> might be better left to a future refactor, just to keep in mind here that we can't get the full metadata object from what we have in the STAC.

schmidtnz (approved):

PR #1313, opened 2025-04-04
fix: existing collection is not updated with new metadata TDE-1444

paulfouquet:
### Motivation

When updating a dataset (added/modified TIFFs for example), the new metadata that is passed to the `collection_from_items.py` script is not used to update the existing Collection.

### Modifications

- Update the existing Collection metadata from the new metadata passed to the `collection_from_items.py` script
- Allow to bypass the `title` re-generation using `--keep-title` flag

>__Note__: If an existing optional metadata field (for example, `--event`) is not passed to the script, the existing metdata field (for example, `linz:event_name`) will be removed.

Example for National DEM:
image

Example not passing the same metadata:
image

### Verification

- Automated tests
- Argo Workflows


scripts/collection_from_items.py
```diff
@@ -103,14 +102,21 @@ def parse_args(args: List[str] | None) -> Namespace:
...
     parser.add_argument(
         "--add-title-suffix",
         dest="add_title_suffix",
-        help="Add a title suffix to the collection title based on the lifecycle. For example, '[TITLE] - Preview'",
+        help="Add a title suffix to the Collection title based on the lifecycle. For example, '[TITLE] - Preview'",
+        type=str_to_bool,
+        required=False,
+    )
+    parser.add_argument(
+        "--keep-title",
```
schmidtnz, line 117: Does this only ever apply to titles or are we going to implement a --keep-X flag for each metadata field?
And if not, should this be more generic (e.g. `--keep` taking a list of fields to keep)?

paulfouquet: At the moment there is only the title that we want to keep in certain cases. I don't see any other metadata that we want to keep. If in the future we need to, then the generic solution would be a better fit.

scripts/stac/imagery/collection.py
```diff
@@ -93,6 +93,44 @@ def from_file(cls, path: str) -> "ImageryCollection":
...
+        if context.lifecycle:
+            self.stac["linz:lifecycle"] = context.lifecycle
+        if context.category:
+            self.stac["linz:geospatial_category"] = context.category
+        if context.region:
+            self.stac["linz:region"] = context.region
+        if context.producers or context.licensors:
+            self.stac["providers"] = []
+            self.add_providers(context.providers)
+        self.stac["description"] = context.description
```
schmidtnz, line 113: does this want the same `if` check as other properties so it can be used to unset / keep description?

scripts/stac/imagery/collection.py
```diff
@@ -93,6 +93,44 @@ def from_file(cls, path: str) -> "ImageryCollection":
...
+        if not keep_title:
```
schmidtnz, line 114: instead of passing an additional flag, would it be sufficient to delete (or set to false-y value) title in context context and treat similar to region, category, etc?
`if context.title:` ... ?

scripts/stac/imagery/collection.py
```diff
@@ -93,6 +93,44 @@ def from_file(cls, path: str) -> "ImageryCollection":
             collection.capture_area = json.loads(read(capture_area_path))
         return collection
 
+    def update(self, context: CollectionContext, updated_datetime: str, keep_title: bool = False) -> None:
+        """Update the Collection with new metadata.
+
+        Args:
+            context: The context containing the updated metadata.
+            updated_datetime: The updated datetime of the Collection.
+            keep_title_desc: Whether to keep the original title and description.
```
schmidtnz, line 102: Nit - outdated docstring (see below before updating this)

scripts/stac/imagery/collection.py
paulfouquet: Do you mean that `CollectionContext` should handle the choice of keeping the existing title rather than in this function?

scripts/stac/imagery/collection.py
paulfouquet: The description is generated from other metadata. We don't have a need to keep the existing description at the moment. I think it's best to regenerate it all the time?

schmidtnz (approved):

PR #1317, opened 2025-04-17
fix: title and description years should match temporal extent TDE-1464

paulfouquet:
### Motivation

The STAC Collection `title` and `description` dates should match its `temporal` `extent` rather than taking the start date and end date of the newly added STAC Items to avoid any issue when re-supplying Items in an existing Collection.

### Modifications

- Use the `temporal` `extent` as dates for the `title` and `description`
- `title` and `description` can't be generated if the Collection does not have a `temporal` `extent` (a Collection needs to have Items in order to get its `title` and `description`)

### Verification

Automated tests


schmidtnz (commented):
This PR reverts some previous changes to a large extent. Was that done intentionally, or is the PR based on an old version?

paulfouquet:
> This PR reverts some previous changes to a large extent. Was that done intentionally, or is the PR based on an old version?

This is intentional. Assuming you're pointing out the title and description generation that moved from `CollectionContext` back to the `ImageryCollection` class. The dates used to generate the title and the descriptions are now coming from the Collection `temporal` `extent` which is calculated during the Collection object lifecycle, instead of being calculated before instantiating the `CollectionContext` previously.
I am happy to any suggestion to make this change smoother.

schmidtnz (approved):

PR #1326, opened 2025-06-11
feat: add dataset domain TDE-1542

paulfouquet:
### Motivation

Some datasets share the same category (for example, "DEM") but diverge on their domain such as the Coastal datasets.

### Modifications

- add a dataset domain to handle Collection Title and Description. Default domain is "Land" and is not visible in the generated data.

See https://github.com/linz/elevation/pull/420

### Verification

Automated tests
Argo Workflows

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -265,20 +265,37 @@ class ImageryCollection:
         Returns:
             Dataset Description
         """
+        IMAGERY = {SCANNED_AERIAL_PHOTOS, SATELLITE_IMAGERY, URBAN_AERIAL_PHOTOS, RURAL_AERIAL_PHOTOS}
+        ELEVATION = {DEM, DSM}
+        HILLSHADES = {DEM_HILLSHADE, DEM_HILLSHADE_IGOR, DSM_HILLSHADE, DSM_HILLSHADE_IGOR}
+
+        category = self.stac["linz:geospatial_category"]
+
+        components = [DATA_DOMAINS[self.domain] if category in {*ELEVATION, *HILLSHADES} else None]
+        if category in {*IMAGERY, *ELEVATION}:
+            components.extend(self._get_imagery_description_components())
+        elif category in HILLSHADES:
+            components.extend(self._get_hillshade_description_components())
+        else:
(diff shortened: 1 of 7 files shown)


scripts/stac/imagery/collection.py
```diff
@@ -283,10 +289,16 @@ def set_description(self) -> None:
             DSM: "Digital Surface Model",
         }
 
+        subtype_prefix = DATA_DOMAINS[self.domain]
+        desc_prefix = ""
+        # domain is only used for DEM/DSM categories
```
schmidtnz, line 294: Can this part of the change be simplified? Maybe by including all categories or refactoring how the description gets assembled?

scripts/stac/imagery/constants.py
```diff
@@ -24,6 +29,11 @@
     DSM_HILLSHADE_IGOR: "DSM Hillshade - Igor",
 }
 
+DATA_DOMAINS = {
```
schmidtnz, line 32: Maybe `DATA_DOMAIN_TITLES` or so? I found it confusing seeing this defined as `DATA_DOMAINS` when the domains are already listed on lines 13-15.

schmidtnz (changes requested):

scripts/stac/imagery/collection.py
schmidtnz: Also, this still refers to `subtype` where everything else was renamed to `domain`

paulfouquet: > Also, this still refers to subtype where everything else was renamed to domain
Has been omitted by the refactor

paulfouquet: > Can this part of the change be simplified? Maybe by including all categories or refactoring how the description gets assembled?

I would not want to include all categories, or would have to do it for the title generation as well. However, it is not in the scope.
I'm opened for a simplification if you have any.

scripts/stac/imagery/constants.py
schmidtnz: HUMAN_READABLE_REGIONS is a good example, I don't like DATA_CATEGORIES and DATA_DOMAINS a lot but don't have anything that would be significantly better

PR #1338, opened 2025-07-14
feat: keep existing collection description TDE-1498

amfage:
### Motivation

:mega: **Note:** this is linked to https://github.com/linz/topo-workflows/pull/1052 which will use this flag.

The DEM merged hillshades need a custom description in the same way that some datasets need a custom title. Add an option to keep existing descriptions (load them from the ODR URL).

### Modifications

Added a new flag `--keep-description` when updating the collection for the merged hillshades.

### Verification

Unit tests and tested with a workflow. The collection was created with the existing description.


paulfouquet (approved):
lgtm

PR #1374, opened 2025-08-20
feat: use supplied capture area TDE-1604

amfage:
### Motivation

For 3D Coastal mapping datasets the capture areas automatically generated when standardising are very complex. This change is to allow a simpler capture area to be added when creating a collection rather than calculating it from the TIFF footprints.

### Modifications

Added a new parameter `supplied_capture_area` to be used by the `create-collection` step.

### Verification

Unit tests.
Ran workflows manually against topo-imagery PR container.

### DIFF
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -392,10 +392,6 @@ class ImageryCollection:
             "file:checksum": file_checksum,
             "file:size": len(capture_area_content),
         }
-        print("==============================")
-        print(supplied_capture_area)
-        print("==============================")
-        print(capture_area)
         self.stac.setdefault("assets", {})["capture_area"] = capture_area
 
         # Save `capture-area.geojson` in target
--- a/scripts/stac/imagery/collection.py
+++ b/scripts/stac/imagery/collection.py
@@ -347,7 +347,9 @@ class ImageryCollection:
 
         return components
 
-    def add_capture_area(self, polygons: list[BaseGeometry], target: str, artifact_target: str = "/tmp") -> None:
+    def add_capture_area(
+        self, polygons: list[BaseGeometry], target: str, supplied_capture_area: str | None, artifact_target: str = "/tmp"
+    ) -> None:
         """Add the capture area of the Collection.
         If the Collection is an update of a published dataset, the existing capture area will be merged with the new one.
         The `href` or path of the capture-area.geojson is always set as the relative `./capture-area.geojson`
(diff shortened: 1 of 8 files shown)


scripts/standardise_validate.py
```diff
@@ -58,6 +58,13 @@ def parse_args() -> argparse.Namespace:
         required=True,
     )
     parser.add_argument("--cutline", dest="cutline", help="Optional cutline to cut imagery to", required=False, nargs="?")
+    parser.add_argument(
```
paulfouquet, line 61: As we have a control on passing the option `--create-footprint`, why not don't pass it when we don't want to create a capture area? That would remove this complexity below.

scripts/stac/imagery/create_stac.py
paulfouquet, line 1: How are managed update of existing datasets regarding this supplied capture area?

paulfouquet: As discussed that's okay

scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -252,9 +260,11 @@ def test_should_make_compliant_capture_area(subtests: SubTests) -> None:
         ),
     ]
 
-    capture_area = merge_polygons(polygons, 0.1)
+    merged_polygons = merge_polygons(polygons, 0.1)
+    mp = cast(MultiPolygon, merged_polygons)
```
paulfouquet, line 264: We might need to do that, to verify

amfage: It was the PyCharm default linter requiring it but Pylint is fine with it so I removed the cast.

scripts/stac/imagery/collection.py
```diff
@@ -375,8 +378,23 @@ def add_capture_area(self, polygons: list[BaseGeometry], target: str, artifact_t
...
-            "description": "Boundary of the total capture area for this collection. Excludes nodata areas in the source "
-            "data. Geometries are simplified.",
+            "description": (
+                "Boundary of the total capture area for this collection"
+                f"{(' provided by the data supplier. '
+                    'May include some areas of nodata where capture was attempted but unsuccessful.')
+                if supplied_capture_area else '. Excludes nodata areas in the source data.'}"
+                " Geometries are simplified."
+            ),
+            #
```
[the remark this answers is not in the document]
amfage: fixed

schmidtnz (approved):

paulfouquet (approved):

PR #1377, opened 2025-08-22
fix: orient the capture area after making geometry valid TDE-1618

amfage:
### Motivation

When working on https://toitutewhenua.atlassian.net/browse/TDE-1604 the capture area generated did not follow the right hand rule. If the `make_valid` step is removed, the orientation is correct.

### Modifications

Do the `make_valid` step before forcing the orientation to follow the right hand winding order rule. I did contemplate an `is_valid` check but `orient` should not make the geometry invalid.

### Verification

Unit tests and a manual test with the larger capture area that had been reversed.

### DIFF
--- a/scripts/stac/imagery/tests/capture_area_test.py
+++ b/scripts/stac/imagery/tests/capture_area_test.py
@@ -4,6 +4,7 @@ from typing import cast
 
 from shapely import get_exterior_ring, is_ccw
 from shapely.geometry import MultiPolygon, Polygon, shape
+from shapely.predicates import is_valid
 
 from scripts.stac.imagery.capture_area import generate_capture_area, merge_polygons, to_feature
 
@@ -150,6 +151,7 @@ def test_capture_area_orientation_polygon() -> None:
     )
     capture_area = generate_capture_area(polygons, Decimal("0.05"))
     assert is_ccw(get_exterior_ring(shape(capture_area["geometry"])))
+    assert is_valid((shape(capture_area["geometry"])))
 
 
 def test_capture_area_orientation_multipolygon() -> None:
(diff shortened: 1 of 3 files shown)


schmidtnz (approved):

scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -229,3 +232,25 @@ def test_capture_area_rounding_decimal_places() -> None:
     }
     capture_area = generate_capture_area(polygons, Decimal("1"))
     assert capture_area == capture_area_expected
+
+
+def test_should_make_valid_capture_area() -> None:
```
paulfouquet, line 237: As you're adding a new verification, this test is not only verifying the geometry is valid, I would change the test name

```suggestion
def test_should_make_compliant_capture_area() -> None:
```

and may be separated the tests using subtests as below

scripts/stac/imagery/tests/capture_area_test.py
```diff
@@ -229,3 +232,25 @@ def test_capture_area_rounding_decimal_places() -> None:
...
+            {
+                "type": "MultiPolygon",
+                "coordinates": [[[[1, 0], [2, 2], [1, 2], [1, 0]]]],
+            }
+        ),
+    ]
+
+    capture_area = merge_polygons(polygons, 0.1)
+    assert is_valid(capture_area)
+    assert is_ccw(get_exterior_ring(capture_area.geoms[0]))
```
paulfouquet, line 256:
```suggestion
    with subtests.test(msg="Valid geometry"):
        assert is_valid(capture_area)

    with subtests.test(msg="Is counterclockwise"):
        assert is_ccw(get_exterior_ring(capture_area.geoms[0]))
```

schmidtnz (approved):

PR #1505, opened 2026-05-14
build: move to uv package manager TDE-1917

amfage:
### Motivation

In preparation for a refactor of this repo, we want to move from Poetry to uv for a faster and more robust package manager.

### Modifications

Moved package files, Dockerfile, GitHub Actions, dependabot and version bump script to uv.
Updated packages.
Small updates related to Black and Pylint.
Moved tests from geos 3.11 compatibility to geos 3.12+.

### Verification

Local build and GitHub Actions.
Tested on PR container to check particularly capture area after Shapely/geos package upgrade.


pyproject.toml
```diff
@@ -1,10 +1,52 @@
...
+    "mypy",
+    "mypy-boto3-s3",
+    "pre-commit",
+    "pylint",
+    "pytest",
+    "pytest-dependency",
+    "pytest-mock",
+    "pytest-subtests",
+    "shellcheck-py",
+]
```
schmidtnz, line 41: "boto3-stubs",
could be a useful dev dependency

.github/workflows/format-tests.yml
```diff
@@ -14,16 +14,18 @@ jobs:
       - uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v4
         with:
           python-version: "3.12.3"
```
schmidtnz: remove this step and add `with` block to `Install uv` to handle py version

schmidtnz:
dependabot needs to migrate to `uv` too

pyproject.toml
amfage: Will note for the future, thank you.

paulfouquet (approved):
LGTM

PR #1572, opened 2026-07-02
refactor: create topo-imagery-stac package TDE-1960

amfage:
### Motivation

We want to update the topo-imagery repo so that we can more easily re-use code modules in separate containers. This is the third of several changes to move code into separate uv packages so that the codebase is easier to work on.

### Modifications

Moved stac-related functions into a separate `topo-imagery-stac` uv package. (Note that the name of the package and the repo will be renamed later once all structural refactoring has been done).

### Verification

Unit tests, end-to-end tests, container build and test standardising workflow run.


ctali (approved):
Reviewed the files with Claude.

PR #1597, opened 2026-07-30
chore: Updated dependabot TDE-1991

ctali:
### Motivation

To implement security recommendations for Dependabot and GitHub Actions
so that we are security compliant

### Modifications

Updated dependabot with cooldown period for every schedule.


amfage (approved):
Looking forward to having this implemented in this noisy repo!

PR #1603, opened 2026-08-04
feat: title generation needs to consistently account for event names TDE-2010

ctali:
### Motivation

To consistently apply event names in titles so that user expectations for titles align across all metadata.

### Modifications
Generated combined name with <geographic_description> or <region> then <event name>, removing duplicated terms.


amfage (approved):
LGTM 👍

PR #1614, opened 2026-08-24
feat: identify 16/32 bit tiffs and write stac metadata accordingly TDE-2004

ctali:
### Motivation

Identify 16-bit and 32-bit TIFFs during STAC item creation and map those GDAL band types into STAC raster metadata.

### Modifications

Updated raster metadata for high bit depth imagery.


packages/topo-imagery-stac/src/topo_imagery_stac/imagery/collection_context.py
```diff
@@ -21,6 +21,7 @@ class CollectionContext:  # pylint:disable=too-many-instance-attributes
         domain (str): The domain of the dataset (e.g., "land").
         region (str): The region of the dataset (e.g., "auckland").
         gsd (Decimal): Ground Sample Distance in meters.
+        data_type (UInt): The data type of the dataset (e.g., "RGB", "DEM").
```
amfage:
```suggestion
        data_type (str): The data type of the dataset (e.g., "uint8", "uint16").
```

.pre-commit-config.yaml
```diff
@@ -39,6 +39,7 @@ repos:
       - id: pylint
         name: pylint
         entry: pylint
+        args: [--disable=too-many-instance-attributes]
```
schmidtnz: do we want this disabled on the whole repo?

packages/topo-imagery-stac/src/topo_imagery_stac/imagery/collection.py
```diff
@@ -117,6 +119,7 @@ def __init__(
             "linz:security_classification": "unclassified",
             "linz:slug": context.linz_slug,
             "gsd": float(context.gsd),
+            "data_type": context.data_type,
```
amfage: Just a note that this has been discussed and will be changed to not write the field if the `data_type` is 8-bit.

schmidtnz: should this be `linz:` namespaced? Or is there a spec / extension we need to include that defines `data_type` as a field?

scripts/collection_from_items.py
```diff
@@ -72,6 +72,7 @@ def get_args_parser() -> CommonArgumentParser:
         choices=HUMAN_READABLE_REGIONS.keys(),
     )
     parser.add_argument("--gsd", dest="gsd", help="GSD of imagery Dataset, for example 0.3", type=str_to_gsd, required=True)
+    parser.add_argument("--data-type", dest="data_type", help="Dataset data type, e.g. uint8", type=str, required=True)
```
schmidtnz: is `required=True` correct if we do not want to track the value when it is `uint8`? 
And with (currently) only 2-3 valid values `uint16`, `uint32`, maybe `uint8` it would be good to provide a list of `choices=`

packages/topo-imagery-stac/src/topo_imagery_stac/imagery/collection.py
amfage: The decision was made to include it at the top level like `gsd`, `data_type` is supported in STAC v1.1.0 although we are not on that version yet.

scripts/collection_from_items.py
schmidtnz:
```suggestion
    parser.add_argument("--data-type", dest="data_type", help="Dataset data type, e.g. uint8", type=str, required=True, choices=['uint8', 'uint16', 'uint32'])
```

amfage: Yeah it should be an optional value with a default of `uint8` so we do not have to specify it in all workflows.

amfage:
superseded by https://github.com/linz/topo-imagery/pull/1641

PR #1626, opened 2026-09-03
refactor: read gdal and pdal versions from bins not env TDE-2048

schmidtnz:
### Motivation

Refactor to stop relying on ENV vars set by the docker_entrypoint.sh shell script, in order to facilitate splitting PDAL and GDAL related scripts into their own containers.

### Modifications

Created helper functions to read version number from pdal and gdal.
Use their output instead of relying on ENV variables.

### Verification

Added unit tests.


packages/topo-imagery-gdal/src/topo_imagery_gdal/gdal/gdal_helper.py
```diff
@@ -99,6 +100,23 @@ def run_gdal(
...
+
+    Raises:
+        GDALExecutionException: if `gdalinfo` cannot be run, or reports no version
+
+    Returns:
+        the `gdalinfo --version` output, for example "GDAL 3.10.3, released 2025/04/01"
+    """
+    version = run_gdal(["gdalinfo", "--version"]).stdout.decode().strip()
+    if not version:  # falsy value raises to prevent empty STAC metadata `processing:software`
+        raise GDALExecutionException("`gdalinfo --version` reported no version")
```
amfage, line 115: Should this have the same error handling as the PDAL equivalent?

schmidtnz: `run_gdal()` includes that error handling already. 
The better approach would be to mirror more closely how `run_gdal()` does this. 
It would be more of a change to `run_pdal()` than I wanted to include here.

amfage (dismissed):

amfage (approved):

PR #1641, opened 2026-09-10
feat!: use data-type parameter to determine BIGTIFF option TDE-2034

amfage:
### Motivation

Use new `data_type` parameter for identifying 16/32-bit data in topo-imagery

### Modifications

Add a `data-type` parameter for simplification as there is no need to read from the tiff because `stac-setup` and `tileindex-validate` are already checking the bit depth.

### Verification

Unit tests, end-to-end tests, local container run.

### DIFF
--- a/packages/topo-imagery-gdal/src/topo_imagery_gdal/gdal/gdal_commands.py
+++ b/packages/topo-imagery-gdal/src/topo_imagery_gdal/gdal/gdal_commands.py
@@ -43,16 +43,19 @@ def get_gdal_command(preset: str, epsg: int, data_type: str = DataType.UINT8.val
     """Build a `gdal_translate` command based on the `preset`, `epsg` code, with conversion to 8bits if required.
 
     Args:
-        preset: gdal preset to use. Defined in `gdal.gdal_preset.py`
+        preset: gdal preset to use. Defined in `gdal.gdal_presets.py`
         epsg: the EPSG code of the file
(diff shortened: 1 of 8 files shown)


schmidtnz:
would be good to get e2e tests asserting both the default (8bit) and non-8bit path (future work?)

packages/topo-imagery-gdal/src/topo_imagery_gdal/gdal/gdal_commands.py
```diff
@@ -38,22 +39,24 @@ def get_buffer_distance(gsd: Decimal) -> float:
     return float(gsd * 2 * DECIMAL_DEGREES_1M)
 
 
-def get_gdal_command(preset: str, epsg: int, band_type: str | None = None) -> list[str]:
+def get_gdal_command(preset: str, epsg: int, data_type: str = DataType.UINT8.value) -> list[str]:
```
schmidtnz: NIT / future work: throughout the PR, I feel like this could be tidied to have data_type as `DataType` ant `str`

amfage: I think save for future work when converting the enums to `strEnum`

schmidtnz (dismissed):

schmidtnz (dismissed):

schmidtnz (approved):

PR #1643, opened 2026-09-13
feat!: add data type to STAC collection metadata TDE-2004

amfage:
@ctali PR changes rebased onto [PR-1641](https://github.com/linz/topo-imagery/pull/1641)

### Motivation & Modifications

Record the dataset data type on the Collection, omitting the key when it is the default uint8. The value comes from the `--data-type` parameter rather than inspecting TIFF bands, consistent with TDE-2034.

Also sets pytest pythonpath so the suite resolves from the repo root.

### Verification

Unit tests and end-to-end tests.


packages/topo-imagery-stac/src/topo_imagery_stac/imagery/collection_context.py
```diff
@@ -40,6 +41,7 @@ class CollectionContext:  # pylint:disable=too-many-instance-attributes
     domain: str
     region: str
     gsd: Decimal
+    data_type: str
```
schmidtnz: `DataType` enum not `str`?

schmidtnz:
```suggestion
    data_type: DataType
```

schmidtnz (dismissed):

schmidtnz (approved):

PR #1651, opened 2026-09-16
feat: tini init container entrypoint

schmidtnz:
### Motivation

Python (and our previous docker-entrypoint.sh script) do not handle signals by default.

### Modifications

Added `tini` entry point to handle incoming signals.

### Verification

Test build, local tests


Dockerfile.pointcloud
```diff
@@ -13,6 +13,9 @@ ENV UV_PROJECT_ENVIRONMENT=/venv
 ENV UV_PYTHON=/usr/bin/python3
 ENV UV_PYTHON_PREFERENCE=only-system
 
+RUN apt-get update \
+    && apt-get install --assume-yes --no-install-recommends tini
```
amfage, line 17: Do we want to pin `tini`?

schmidtnz: main reason to not pin was because from what I understand the versions don't stick around very long in ubuntu images / apt mirrors. And dependabot doesn't manage those version numbers.

amfage (dismissed):

amfage (approved):

PR #1653, opened 2026-09-16
fix: stream file copies to allow outputs over 5GB TDE-2072

schmidtnz:
### Motivation

16-bit COGs may exceed 5GB, which `PutObject` rejects with `EntityTooLarge`.

### Modifications

`fs.copy()` streams instead of buffering, using `upload_file` for `s3` so multipart uploads are handled automatically. It now returns the multihash rather than the target path. Call sites doing `write(target, read(source))` moved to it.

STAC reuses the multihash instead of reading the asset back to checksum it.

### Verification

Unit tests for the streaming helpers, all four `copy()` directions, and the checksum passthrough; `file:checksum` values pinned against precomputed hashes.

IN PROGRESS: Mackenzie RGBNIR workflow [test run](https://argo.linzaccess.com/workflows/argo/is-mackenzie-rgbnir-pr1653-57zj5?tab=workflow&nodeId=is-mackenzie-rgbnir-pr1653-57zj5-806484618&uid=2140bc57-836f-452f-abb7-50eb1de28240) to confirm the >5GB upload end to end.


packages/geoprocessor-common/src/geoprocessor_common/files/fs.py
```diff
@@ -50,18 +74,51 @@ def read(path: str) -> bytes:
...
+            fs_local.copy_file(source, target)
+            file_multihash = fs_local.multihash(target)
+
+        get_log().debug(
+            "copy_success", source=source, target=target, multihash=file_multihash, duration=time_in_ms() - start_time
+        )
+        return file_multihash
+    except FileNotFoundError as error:
+        raise NoSuchFileError(source) from error
+    except client("s3").exceptions.ClientError as ce:
```
ctali: could the exception be handled without creating another S3 client here?

packages/geoprocessor-common/src/geoprocessor_common/files/fs_s3.py
```diff
@@ -78,13 +117,66 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
...
+    Args:
+        path: The AWS S3 path to the file to read.
+        destination: The local path to the file to write.
+        needs_credentials: Tells if credentials are needed. Defaults to False.
+
+    Raises:
+        ClientError
+    """
+    start_time = time_in_ms()
+    os.makedirs(os.path.dirname(destination), mode=0o777, exist_ok=True)
```
ctali: Or is destination always expected to have a directory? Otherwise, could os.path.dirname(destination) return an empty string and cause os.makedirs() to fail?

schmidtnz: fixed

packages/geoprocessor-common/src/geoprocessor_common/files/fs.py
schmidtnz: fixed (also in pre-existing places)

packages/geoprocessor-common/src/geoprocessor_common/files/fs_s3.py
```diff
@@ -67,24 +107,79 @@ def read(path: str, needs_credentials: bool = False) -> bytes:
...
+    Args:
+        path: The AWS S3 path to the file to read.
+        destination: The local path to the file to write.
+        needs_credentials: Tells if credentials are needed. Defaults to False.
+
+    Raises:
+        ClientError
+    """
+    start_time = time_in_ms()
+    with fs_local.atomic_write_path(destination) as partial_destination:
```
amfage, line 157: Will this leave an empty file behind if `_get_object_body()` fails?

schmidtnz: The `.part` file might remain if it fails but not the actual `destination` (and it won't truncate `destination` for the same reason).

amfage (approved):

